Issue #1, opened 2021-02-10
postgresql.enabled: false still spins up postgresql


iandvt:
# reproduction steps

## install
``` bash
helm repo add retool https://charts.retool.com
kubectl create namespace retool
helm -n retool install retool retool/retool -f retool.yaml
```
## retool.yaml
``` yaml
config:
  licenseKey: LOCAL-ONLY-TRIAL

  postgresql:
    host: retool-helm.postgres.database.azure.com
    port: 5432
    db: postgres
    user: retool@retool-helm
    password: *REDACTED*
    ssl_enabled: true

postgresql:
  enabled: false

```

## verification
```
kubectl -n retool get pods
```

<img width="650" alt="Screen Shot 2021-02-09 at 9 14 06 PM" src="https://user-images.githubusercontent.com/25870869/107467883-d6627d80-6b1b-11eb-85a1-fa1863f8d9fd.png">

## cleanup
```
helm -n retool delete retool
```

philippeboyd:
@iandvt you shouldn't put credentials on a github issue and should change your postgresql password right now...

iandvt:
@philippeboyd - I deployed and left this available so the developers had something to debug/test against. I probably should have left a note in the original post. That being said, I have decommissioned this Azure Postgres database.

kyle-retool:
@iandvt thanks for helping out. I test this chart locally using Docker Desktop, so won't need any public resources.

PR #4, opened 2021-02-10
fix: missing postgresql chart condition

philippeboyd:
fixes https://github.com/tryretool/retool-helm/issues/1


kyle-retool:
Thanks!

PR #5, opened 2021-02-10
fix: add missing service account file

philippeboyd:
also changed the `create` value from `false` to `true`.

### DIFF
--- a/templates/serviceaccount.yaml
+++ b/templates/serviceaccount.yaml
@@ -1,0 +1,10 @@
+{{- if .Values.serviceAccount.create }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "retool.serviceAccountName" . }}
+{{- if .Values.serviceAccount.annotations }}
+  annotations:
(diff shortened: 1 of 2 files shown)


kyle-retool (approved):

kyle-retool:
Should `create` be renamed to `enabled`?

philippeboyd:
No because it'll imply that a service account is optional where in Kubernetes a service account is always used. The use case is more: do you want to create a service account or use an existing one, therefore `create: true|false` is appropriated.

In most cases, you'll always want to create a service account specific to your application.

Other renowned helm charts do this such as [Vault](https://github.com/hashicorp/vault-helm/blob/84a1dd9fbe00fa7658c8165cbc7f9bc82a2fba5d/values.yaml#L589) or [Ambassador](https://github.com/datawire/ambassador-chart/blob/f2385fece4f94b77be91c311264579bc4a855b62/values.yaml#L234)

kyle-retool:
Published version 4.1.1 with this change.

Issue #6, opened 2021-02-10
Cannot bump Retool version


kentwalters:
With this Helm chart we can’t currently bump the Retool version as it is specified in the chart itself and not an editable value.

kyle-retool:
> and not an editable value

I believe that this actually is a configurable value. In `values.yaml`:

```yaml
image:
  repository: "tryretool/backend"
  tag: "latest"
  pullPolicy: "IfNotPresent"
```

So I think if you add the following to the values file passed into `helm install`, you should be able to set the version.

```yaml
image:
  tag: "2.66.14"
```

philippeboyd:
@kyle-retool better yet: you can set the tag in the deployments' yaml to be `{{ default .Chart.AppVersion .Values.image.tag }}` instead of `{{ .Values.image.tag }}`

That way it will default to the Chart's appVersion if people don't specify an image tag

philippeboyd:
Just opened a PR for you guys

PR #7, opened 2021-02-10
feature: default to chart appVersion if image tag is empty

philippeboyd:
Fixes https://github.com/tryretool/retool-helm/issues/6


kyle-retool:
Excellent! Thanks again.

PR #9, opened 2021-04-01
[helm2] Change the latest tag to a defined version number

apurvas:
In this PR, @jssln and I are changing the previous "latest" tag to a defined version number. The reason for this is to ensure there are no inconsistencies with upgrades to different version numbers. Because we have two containers, it's possible that one of the containers will have a different version due to the underlying "latest" changing. 

### Test Plan:

Deployed our own on-premise build to confirm installation is smooth and works. 
Screen Shot 2021-04-03 at 9 59 53 PM

We will also add updated documentation in our README regarding the image.tag value and what it means to change it. 
* PR for the README change: https://github.com/tryretool/retool-helm/pull/10

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -22,7 +22,7 @@ config:
 
 image:
   repository: "tryretool/backend"
-  tag: "2.66.65"
+  tag: "X.Y.Z"
   pullPolicy: "IfNotPresent"
 
 commandline:
--- a/values.yaml
+++ b/values.yaml
@@ -22,7 +22,7 @@ config:
 
 image:
   repository: "tryretool/backend"
-  tag: "latest"
+  tag: "2.66.65"
   pullPolicy: "IfNotPresent"
 
 commandline:


values.yaml
```diff
@@ -22,7 +22,7 @@ config:
 
 image:
   repository: "tryretool/backend"
-  tag: "latest"
+  tag: "2.66.65"
```
jssln, line 25: As I was writing the README update, I thought, could we set this to `X.XX.XX`? This would make the user deliberately set a version.

It would help specify what new or existing users what to put here, and prevent the problem we were talking about (i.e. existing users merging this change, not noticing the version change, and unwittingly getting a different version than they expected).

What do you think @apurvas?

apurvas, line 25:
```suggestion
  tag: "X.Y.Z"
```

Discussed offline. Let's go with this.

apurvas: Discussed offline. This sounds good.

jssln (approved):

PR #10, opened 2021-04-06
Update Helm 2 README

jssln:
* Explicitly call out that you should specify a Retool version, and add links to help the user figure out the version they want.
* Pull the instructions for specifying the Postgres DB details into the set up steps.

Will make a similar change for the Helm 3 README, after getting review on this one.


apurvas (approved):
This is great! Let's also (maybe another PR) add in a bit about updating Helm 2 as well

PR #13, opened 2021-04-12
Update Helm 3 readme with instructions to set `encryptionKey` and `jwtSecret`

jssln:
Motivation: 
@alexwestreich points out that it's vital to set these two env vars. We can update the Environment Variables page itself to give more color on what changing these values will lead to.


apurvas (approved):

alexwestreich:
updated the docs: 
image

PR #15, opened 2021-04-29
Allow more granular control for secrets

rhinon:
As it is, it's somewhat annoying to use some secrets and not others since the current mechanism to use external secrets is all-or-nothing with a whole block of env vars.

This allows you to optionally use:
1. `config.licenseKeySecretName` instead of `config.licenseKey`
2. `config.jwtSecretSecretName` instead of `config. jwtSecret`
2. `config.encryptionKeySecretName` instead of `config.encryptionKey`
3. `config.auth.google.clientSecretSecretName` instead of `config.auth.google.clientSecret`
4. `config.postgresql.passwordSecretName` instead of `config.postgresql.password`

In all of these instances, the alternative `<x>SecretName` allows you to specify the name of the k8s secret as an alternative to hardcoding the secrets into values and having the chart put them in secrets.

### DIFF
--- a/templates/deployment_backend.yaml
+++ b/templates/deployment_backend.yaml
@@ -67,17 +67,29 @@ spec:
           - name: LICENSE_KEY
             valueFrom:
               secretKeyRef:
+                {{- if .Values.config.licenseKeySecretName }}
+                name: {{ .Values.config.licenseKeySecretName }}
+                {{- else }}
                 name: {{ template "retool.fullname" . }}
+                {{- end }}
                 key: license-key
           - name: JWT_SECRET
             valueFrom:
               secretKeyRef:
+                {{- if .Values.config.jwtSecretName }}
+                name: {{ .Values.config.jwtSecretName }}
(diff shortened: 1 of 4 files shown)


templates/deployment_backend.yaml
```diff
@@ -67,17 +67,29 @@ spec:
           - name: LICENSE_KEY
             valueFrom:
               secretKeyRef:
+                {{- if .Values.config.licenseKeySecretName }}
+                name: {{ .Values.config.licenseKeySecretName }}
+                {{- else }}
                 name: {{ template "retool.fullname" . }}
+                {{- end }}
                 key: license-key
```
rhinon, line 75: should I change the keys to camelcase? It will be harder to change later since this is not currently exposed right now as is.

rhinon: Changed this to be configurable

kyle-retool (approved):
Looks good to me.

PR #16, opened 2021-05-03
Fix missing usage of postgresql.ssl_enabled in jobs deployment

chamel-shipbob:
Azure Postgres requires SSL to communicate.
The flag exist in the retool server deployment but not in the backend job deployment.
This causes upgrades to fail as the backend job is unable to complete migrations against postgres.
This fix also includes support for liveness failure threshold so that it can be modified


rhinon (approved):
Thank you! Looks good to me.

PR #17, opened 2021-05-03
Remove appVersion from chart

rhinon:
It doesn't make sense for us to declare a static version here as this chart is not released per new version of Retool. Given our updated instructions (https://github.com/tryretool/retool-helm/pull/11) to specify a Retool version, we should just remove this.

Per Helm, it is optional: https://helm.sh/docs/topics/charts/#the-appversion-field

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -3,7 +3,6 @@ name: retool
 description: A Helm chart for Kubernetes
 type: application
 version: 4.3.0
-appVersion: "2.66.14"
 maintainers:
   - name: Retool Engineering
     email: engineering+helm@retool.com


kyle-retool:
I actually disagree with removing this. I think we should instead update the Helm chart every time we release a new stable version, which is once every two weeks.

rhinon:
> I actually disagree with removing this. I think we should instead update the Helm chart every time we release a new stable version, which is once every two weeks.

Don't mind deferring, but until we actually do this, our appVersion is meaningless. Is it the best practice to release a new helm chart for every new version of the app, even if the chart itself does not change?

PR #21, opened 2021-07-31
Skip unnecessary secret generation

rhinon:
If you use `.Values.config.jwtSecretSecretName` and `.Values.config.encryptionKeySecretName` instead of specifying one in your values file, the helm chart will still generate random secrets for you. 

This is usually fine because of helm resource policy (`no-upgrade-existing`), but the terraform helm provider detects this as the provider producing inconsistent terraform plans. So this changes the secrets to not generate when those config values are used. Otherwise the helm provider thinks it is generating new random secrets on every terraform plan and terraform apply and never works.

### DIFF
--- a/templates/secret.yaml
+++ b/templates/secret.yaml
@@ -13,16 +13,20 @@ type: Opaque
 data:
   license-key: {{ .Values.config.licenseKey | default "" | b64enc | quote }}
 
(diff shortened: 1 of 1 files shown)


anna-yn (approved):

mayzenator (approved):

kyle-retool:
I'll make sure to release a new version of the Helm chart today

PR #22, opened 2021-08-04
Fix incorrect references to jwtSecretSecretName

rhinon:
🤦 The correct value from the values file is `jwtSecretSecretName`, not `jwtSecretName`


kyle-retool (approved):
YAML programming strikes again

PR #33, opened 2021-11-26
Fixup: remove references to AppVersion

Pluies:
Hey @rhinon @kyle-retool ,

I see we've removed the default to AppVersion in #20 (with discussion in #17).

This PR cleans up a few loose ends from that PR:
- Removes the `default`, and makes `.Values.image.tag` mandatory
- Remove the outdated comment about defaulting to AppVersion

Feel free to close this PR if the goal is to come back to using AppVersion as alluded to in #17 !

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 4.6.0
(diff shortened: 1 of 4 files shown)


anna-yn (approved):
@Pluies sorry for the late reply, thank you so much for the PR!

Issue #34, opened 2021-12-03
Add escape hatch for extra manifests and support for environment variables from secrets


gab-nando:
It would be extremely handy to have the possibility to:
* Add arbitrary manifests to the chart. One use case is to be able to provision ManagedCertificate, BackendConfig to be used with Google Load Balancer.
* Set environment variables from secret. Helpful, for instance, for setting SCIM_AUTH_TOKEN

stpierre:
Yes please! We'd like to add HPAs, and the inability to pass in env vars from secrets is a pretty substantial oversight. Hopefully this gets attention quickly.

rjay98:
apologies for accidentally assigning - @kyle-retool if you want to take a look!

PR #35, opened 2021-12-03
Add escape hatch for extra manifests + support for env vars from secrets

gab-nando:
Fixes https://github.com/tryretool/retool-helm/issues/34


templates/deployment_backend.yaml
```diff
@@ -132,6 +132,13 @@ spec:
           - name: "{{ $key }}"
             value: "{{ $value }}"
           {{- end }}
+          {{- range .Values.environmentSecrets }}
+          - name: {{ .name }}
+            valueFrom:
+              secretKeyRef:
+                name: {{ .secretKeyRef.name }}
+                key: {{ .secretKeyRef.key }}
+          {{- end }}
```
stpierre, line 141: I typically see this handled with just a singmple `toYaml`:

```
          env:
            {{- toYaml .Values.application.environment | nindent 12 }}
```

It would be a backwards-incompatible change, but then this chart would instantly handle all the current and future complexity of env var descriptions. Not sure how the maintainers would feel about that, though. I want this feature far more than I want this feature to be perfect. :)

MischaWeerwag:
We really need this for our gitops workflow too.

templates/deployment_backend.yaml
anna-yn: yeaa I agree the single `toYaml` is much nicer, but we need to think more about breaking changes 😬  let's get your changes landed first!

values.yaml
```diff
@@ -51,6 +51,10 @@ commandline:
 
 env: {}
 
+# Optionally specify additional environment variables to be populated from Kubernetes secrets.
+# Useful for passing in SCIM_AUTH_TOKEN or other secret environment variables from Kubernetes secrets.
+environmentSecrets: []
```
anna-yn, line 56: nit: can you add an example here?

anna-yn (approved):
just 1 nit, thank you so much for the PR ❤️

values.yaml
gab-nando: Sure :)

gab-nando:
@anna-yn it should be good to go now. If you prefer bumping only the minor version of the chart (4.6.1 -> 4.6.2) instead of releasing 4.7.0, let me know and I'll update it accordingly

anna-yn:
> @anna-yn it should be good to go now. If you prefer bumping only the minor version of the chart (4.6.1 -> 4.6.2) instead of releasing 4.7.0, let me know and I'll update it accordingly

I think 4.7.0 is fine, there's something else we need to land soon, we'll just lump them into one minor release

PR #36, opened 2021-12-27
created ingress for kubernetes v1.22 compatible

prashant637:
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122

### DIFF
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -16,9 +16,12 @@ spec:
   rules:
   - http:
       paths:
-      - backend:
-          serviceName: {{ template "retool.fullname" . }}
-          servicePort: {{ .Values.service.externalPort }}
+      - pathType: {{ .Values.ingress.pathType }}
+        backend:
+          service:
+            name: {{ template "retool.fullname" . }}
+            port: 
+              number: {{ .Values.service.externalPort }}
 {{- if .Values.ingress.hostName }}
     host: {{ .Values.ingress.hostName | quote }}
 {{- end }}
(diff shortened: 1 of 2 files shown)


mukul-c2fo (commented):
@prashant637 - Just wondering if this change would let the chart fail on a lower k8s version. What do you think? If so, we can add some conditionals to apply the ingress based on the k8s version it's going to be deployed to.

prashant637:
@mukul-c2fo Changes has been done to support the old and new ingress.

mukul-c2fo:
@theodormarcu - Could we please get some eyes on this PR as this could be a blocker for us to rollout 1.22.

anna-yn:
Hi @prashant637 @mukul-c2fo , sorry for the late reply and thanks for the PR! Looks like this PR is very similar to #38 , except that this one sets the `ingressClassName` and the other one doesn't? I'd prefer we either don't set the `ingressClassName` for now or make it optional by default (only set it if `ingress.className` has a value, and don't give it a value by default) because otherwise it breaks `helm ugprade` for people who've been setting the ingress class with the `kubernetes.io/ingress.class` annotation. What do you think?

anna-yn:
Hi again! I merged #38 because that one doesn't contain the `ingressClassName` change and it won't affect people's `helm upgrade`. It's in the release v4.7.0 now

prashant637:
Hi @anna-yn 
Sure, we can skip the ingressClassName and can run without using the same. I ran the release v4.7.0, it's working fine. 
Thanks for merging the PR #38

PR #37, opened 2022-01-11
add option to create secret

mirobertod:
Add option to choose whether create the secret or not.

Currently, even if the secret name is not empty (`licenseKeySecretName`, `encryptionKeySecretName`, ...) the secret is always created.
Moreover, if you provide a secret name equals to `{{ template "retool.fullname" . }}` you'll get the error from Helm `rendered manifests contain a resource that already exists`

Thanks


mirobertod:
I close this PR, I think my use case can be solved with `externalSecrets` (https://github.com/tryretool/retool-helm/blob/main/values.yaml#L55)

PR #38, opened 2022-01-11
updated ingress object api version to support k8s ingress v1.22

prashant637:


anna-yn (approved):
I'm going to approve this one instead of #36 since this one is backwards compatible with how the class name is set!

prashant637:
sure @anna-yn

PR #44, opened 2022-03-15
Allow more than one hostname for ingress

stpierre:
This also makes it possible to specify a path; currently setting a
path for the ingress is broken, as it uses `{{ .path }}`, which
appears to have been copied from the default ingress template (as
created by `helm init`) without the `range` that would make that
possible.


anna-yn:
Hi @stpierre , thank you so much for your contribution! I'll test this out and release it in v4.9.0

Issue #41, opened 2022-03-03
Ingress path seems to be a bug


bipinu:
Where does the `{{ .path }}` value in [ingress.yaml](https://github.com/tryretool/retool-helm/blob/main/templates/ingress.yaml#L25
) come from?   
Bases on [NOTES.txt](https://github.com/tryretool/retool-helm/blob/main/templates/NOTES.txt#L3), it seems that Values.yaml should have an array of hosts and paths, but [that doesn't seem to be the case](https://github.com/tryretool/retool-helm/blob/main/values.yaml#L92).

hafeyang:
same problem

anna-yn:
Hi! Sorry for the late reply, this should be fixed now by https://github.com/tryretool/retool-helm/commit/81734ea759431f6ce4ee0fa9ce4f8847bd56ebb0 in v4.9.0!

PR #47, opened 2022-03-27
fix ingress path value

ruckc:
The ingress {{ .path }} template is invalid.  This PR fixes it and exposes an `ingress.path` value to correspond with it.  This addresses #41.

### DIFF
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -22,7 +22,7 @@ spec:
   rules:
   - http:
       paths:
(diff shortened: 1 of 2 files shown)


anna-yn:
Hi @ruckc , thank you so much for your contribution! There's another PR https://github.com/tryretool/retool-helm/pull/44 that addresses this issue and also adds support for multiple paths, so I'll be merging that one instead, but again thank you for putting this up!

PR #51, opened 2022-04-13
Add support for modern external secrets operator

moogzy:
Add support for the modern ExternalSecretsOperator which supports a number of cloud based backends for secrets management/retrieval.

This change allows end users that have the ExternalSecretsOperator (ESO) running in their cluster to manage retool config through secrets. This is especially helpful for driving additional env vars such as the GITHUB env vars related to application security.

I've purposely left the previous/legacy way of defining secrets for backwards compatibility as this doesn't need to be a breaking change. The end users can decide upon whether they continue to use the legacy method plus retool now has the ability to support clusters that are using the modern ESO.

I've tested this locally with our retool helmfiles and it renders correctly and does not apply the config natively as env vars but rather is reading them from a secrets file. This is WAI.

```
+         envFrom:
+         - secretRef:
+             name: retool-config
+         - secretRef:
+             name: retool-db
```

Example of what this looks like ExternalSecret object wise for a AWS Secrets Manager Backend:

**ExternalSecrets Objects**
```
retool, retool-config, ExternalSecret (kubernetes-client.io) has been added:
- 
+ # Source: retool/templates/externalsecret.yaml
+ apiVersion: kubernetes-client.io/v1
+ kind: ExternalSecret
+ metadata:
+   annotations:
+     "helm.sh/hook-weight": "1"
+     "helm.sh/hook": pre-install,pre-upgrade
+   name: retool-config
+   namespace: retool
+ spec:
+   backendType: secretsManager
+   dataFrom:
+     - global/retool
retool, retool-db, ExternalSecret (kubernetes-client.io) has been added:
- 
+ # Source: retool/templates/externalsecret.yaml
+ apiVersion: kubernetes-client.io/v1
+ kind: ExternalSecret
+ metadata:
+   annotations:
+     "helm.sh/hook-weight": "1"
+     "helm.sh/hook": pre-install,pre-upgrade
+   name: retool-db
+   namespace: retool
+ spec:
+   backendType: secretsManager
+   dataFrom:
+     - global/retool/db

```


moogzy:
@anna-yn hi there, would you be in a position to review this patch please?

anna-yn:
> @anna-yn hi there, would you be in a position to review this patch please?

Hey! Thanks for the ping. I'll try to take a look at this today, if not I'll get to it early next week

values.yaml
```diff
@@ -63,10 +63,25 @@ environmentSecrets: []
...
-# Support for external secrets https://github.com/godaddy/kubernetes-external-secrets
+# Enables support for the legacy external secrets (enabled) and the modern External Secrets Operator (externalSecretsOperator.enabled).
+# These are mutually exclusive as both enable reading in environments variables via External Secrets.
 externalSecrets:
+  # Support for legacy external secrets, note this is deprecated in favour of External Secrets Operator: https://github.com/godaddy/kubernetes-external-secrets
+  # This mode only allows a single secret name to be provided.
   enabled: false
   name: retool-config
+  # Support for External Secrets Operator: https://github.com/external-secrets/external-secrets
+  externalSecretsOperator:
```
anna-yn, line 74: do you think it might be confusing to some users that while this is nested inside `externalSecrets`, if they prefer using this approach they should be setting `enabled: false` in `externalSecrets` and the nested `enabled` below to `true`?

moogzy: I think the descriptions provided in the values.yaml file clearly explaining they are mutually exclusive helps clarify that.

What would help even more so is to provide examples in the retool documentation.

I don't see a way where maintaining backwards compatibility is possible without needing to set "enabled: false" for the legacy mode. As even if "externalSecretsOperator" was not nested the boolean for enabled will still be the controlling factor for mutual exclusion.

I think a future PR can deprecate the legacy mode as a potential breaking change in favour of having the modern ESO. WDYT?

anna-yn: ok makes sense!

anna-yn (approved):
thanks again for your contribution! I'll put this in the next release

moogzy:
> thanks again for your contribution! I'll put this in the next release

Thanks Anna. Looks like you're on a monthly cadence for releases, correct? Will this make the May cut?

PR #55, opened 2022-05-26
Add Values.extraEnvs

lsovern:
Add support for env variables that aren't key-value with a new variable called `extraEnvs`. This will make the existing `env` and `environmentSecrets` redundant, but leaving them in so as not to break existing things.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -65,7 +65,7 @@ environmentSecrets: []
 
 # Optionally specify environmental variables. Useful for variables that are not key-value, as env: {} above requires.
 # Can also include environment secrets here instead of in environmentSecrets
-extraEnvs: []
+environmentVariables: []
   #   - name: SCIM_AUTH_TOKEN
   #     valueFrom:
   #       secretKeyRef:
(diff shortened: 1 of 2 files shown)


robert-burden (approved):

values.yaml
```diff
@@ -63,6 +63,24 @@ environmentSecrets: []
...
+  #     valueFrom:
+  #       secretKeyRef:
+  #         name: retool-scim-auth-token
+  #         key: auth-token
+  #   - name: GITHUB_APP_PRIVATE_KEY
+  #     valueFrom:
+  #       secretKeyRef:
+  #         name: retool-github-app-private-key
+  #         key: private-key
+  #   - name: NAME
```
goodspark, line 79: Nitpick:
```suggestion
  #   - name: POD_HOST_IP
```

That's not the exact name we'd use, but it better clarifies the example I think.

prashantmital:
@lsovern can we use something more generic than `extraEnvs`? I agree that this will eventually replace `env` and `environmentSecrets` but to avoid breaking things when we do, I think using a more generic term like `environmentVariables` might be better?

anna-yn (approved):
agree with @goodspark 's suggestion, otherwise lgtm!

prashantmital (approved):

PR #60, opened 2022-06-14
Fix indentation of environmentVariables

goodspark:
When using environmentVariables, the template was being rendered like
this:
```yaml
          - name: CLIENT_SECRET
            valueFrom:
              secretKeyRef:
                name: retool
                key: google-client-secret
                    - name: asd
            value: asd
          - name: zxc
            valueFrom:
              fieldRef:
                fieldKey: asd
```

Notice how the indentation for 'asd' is wrong.

### DIFF
--- a/templates/deployment_backend.yaml
+++ b/templates/deployment_backend.yaml
@@ -146,7 +146,7 @@ spec:
                 key: {{ .secretKeyRef.key }}
           {{- end }}
           {{- with .Values.environmentVariables }}
-          {{ toYaml . | indent 10 }}
+{{ toYaml . | indent 10 }}
           {{- end }}
         {{- if .Values.externalSecrets.enabled }}
         envFrom:


templates/deployment_backend.yaml
```diff
@@ -146,7 +146,7 @@ spec:
                 key: {{ .secretKeyRef.key }}
           {{- end }}
           {{- with .Values.environmentVariables }}
-          {{ toYaml . | indent 10 }}
+{{ toYaml . | indent 10 }}
```
goodspark, line 149: After `helm template retool . -f test-values.yaml`
```yml
          - name: CLIENT_SECRET
            valueFrom:
              secretKeyRef:
                name: retool
                key: google-client-secret
          - name: asd
            value: asd
          - name: zxc
            valueFrom:
              fieldRef:
                fieldKey: asd
```

goodspark: Unfortunately the indentation is needed because toYAML isn't smart enough to detect and prefix current indentation. So once you hit a newline, the subsequent lines lose the indentation, which generates invalid YAML or invalid k8s manifests:
```yaml
          - name: CLIENT_SECRET
            valueFrom:
              secretKeyRef:
                name: retool
                key: google-client-secret
          - name: asd
  value: asd
- name: zxc
  valueFrom:
    fieldRef:
      fieldKey: asd
```

For reference, this is already an established pattern elsewhere in this file: https://github.com/tryretool/retool-helm/blob/main/templates/deployment_backend.yaml#L187

anna-yn: ah ok! makes sense

anna-yn (approved):

PR #65, opened 2022-07-20
Retool helm chart reads the previous encryption key and jwt secret if…

JatinNanda:
Reads from the existing autogenerated encryptionKey and jwtSecret if a specific place for them or new values for them aren't specified

to accomplish this, i'm using the lookup function which seems to be the community-agreed-on workaround: https://github.com/helm/helm/issues/3053

the syntax here was kind of confusing because i had to 
1) get the secret name to respect the template we were using
2) the names of the keys themselves were hyphenated and break with standard templating so i had to use 'include' to get around that

### DIFF
--- a/templates/secret.yaml
+++ b/templates/secret.yaml
@@ -10,13 +10,17 @@ metadata:
     "helm.sh/resource-policy": no-upgrade-existing
 {{- end }}
 type: Opaque
+{{- $secretName := (include "retool.fullname" .) }}
+{{- $secret := lookup "v1" "Secret" .Release.Namespace $secretName }}
 data:
   license-key: {{ .Values.config.licenseKey | default "" | b64enc | quote }}
 
   {{ if not .Values.config.jwtSecretSecretName }}
   {{ if .Values.config.jwtSecret }}
   jwt-secret: {{ .Values.config.jwtSecret | b64enc | quote }}
-  {{ else  }}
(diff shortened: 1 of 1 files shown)


templates/secret.yaml
```diff
@@ -10,21 +10,27 @@ metadata:
...
+{{- $secretName := (include "retool.fullname" .) }}
+{{- $secret := lookup "v1" "Secret" .Release.Namespace $secretName }}
 data:
   license-key: {{ .Values.config.licenseKey | default "" | b64enc | quote }}
 
   {{ if not .Values.config.jwtSecretSecretName }}
   {{ if .Values.config.jwtSecret }}
   jwt-secret: {{ .Values.config.jwtSecret | b64enc | quote }}
-  {{ else  }}
+  {{ else if and $secret (index $secret.data "jwt-secret") }}
```
linnerissa, line 21: why is and needed in this syntax :O

linnerissa (approved):
VERY NICE

templates/secret.yaml
JatinNanda: i think these are like the go functions where and/or operate on the next two arguments. i was using it here as a null check, so like if the secret exists and it has a data field - i was getting some nil complaints on `helm lint` without this but there might be a better way to do it

anna-yn (approved):

PR #67, opened 2022-07-21
add gitignore

TumuGuskun:


JatinNanda (approved):
looks like other charts also don't check this in so yeah seems fine https://github.com/kubernetes/ingress-nginx/tree/main/charts/ingress-nginx

PR #68, opened 2022-08-04
update secret generation for new installs

TumuGuskun:


templates/secret.yaml
```diff
@@ -18,7 +18,7 @@ data:
   {{ if not .Values.config.jwtSecretSecretName }}
   {{ if .Values.config.jwtSecret }}
   jwt-secret: {{ .Values.config.jwtSecret | b64enc | quote }}
-  {{ else if and $secret (index $secret.data "jwt-secret") }}
+  {{ else if and $secret $secret.data (index $secret.data "jwt-secret") }}
```
JatinNanda, line 21: oh you can just string as many things as you want here? did you test this, i guess at least in the happy cases that we _can_ repro?

TumuGuskun: as far as I can tell yeah. I can run some tests. Is there any way we can test it against something that doesn't have a `$secret`

JatinNanda: no fresh installs will have a $secret, but you can create the necessary state you need by just kubectl'ing a secret that matches the name we're looking for and just doesn't have a data block to test this specific thing. 

in terms of reproing what we've been seeing, I think we can try duping a current nameOverride when we launch the instance, i think at least one repro was because of that. 

i feel like the bug must be in how we're generating this secret name, we might be pointing to something that isn't the secret we're looking for somehow (worth seeing if we can repro imo)

PR #73, opened 2022-10-24
Update cpu & memory resource recommendation

lefttree:
For context of making this change, please see [Jobs runner instability research](https://docs.google.com/document/d/1N8Ahnpz5sLUQFX7yJYoyrjN7x37rwqTy6nnbR-FkBXU/edit)

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -200,14 +200,14 @@ extraVolumes: []
 resources: {}
   # If you have more than 1 replica, the minimum recommended resources configuration is as follows:
   # - cpu: 2048m
-  # - memory: 4098Mi
+  # - memory: 4096Mi
   # If you only have 1 replica, please double the above numbers.
   # limits:
-  #   cpu: 500m
-  #   memory: 500Mi
+  #   cpu: 4096m
+  #   memory: 8192Mi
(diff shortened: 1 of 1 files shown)


values.yaml
```diff
@@ -198,6 +198,10 @@ extraVolumeMounts: []
 extraVolumes: []
 
 resources: {}
+  # If you have more than 1 replica, the minimum recommended resources configurate are as follows:
```
JatinNanda, line 201: configurate are as follows: -> configuration is as follows:?

values.yaml
```diff
@@ -198,6 +198,10 @@ extraVolumeMounts: []
...
+  # - cpu: 2048m
+  # - memory: 4098Mi
```
JatinNanda, line 203: it's unclear whether these are requests or limits - also if this is what we suggest, can we just modify the existing 500m 500mi comments below?

lefttree: These should be requests. I updated the following requests and doubled the limit on top of that

JatinNanda (approved):
seems fine, but I'm worried this is a huge jump - could you tldr me how you arrived at these numbers?

PR #76, opened 2022-12-07
Include DB_SSH_CONNECTOR when we scale up replica counts

JatinNanda:
seems like our intention when we increase replica count is just to split out the jobs runner container, meaning we have to specify SERVICE_TYPES explicitly.

looks like we were never specifying the one for the ssh dbconnector, meaning on multi-replica deployments, it'll always be turned off lmao


mayzenator (approved):
lmao

PR #78, opened 2023-01-19
[DTC-3268] codify the resource request for jobs runner and make it a separate pod by default

lefttree:
1. It could avoid customers setting low resources without checking our recommendation
2. Separate jobs runner pod by default would make it easier for the liveness detection https://github.com/tryretool/retool_development/pull/18728

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -197,17 +197,17 @@ extraVolumeMounts: []
 
 extraVolumes: []
 
-resources: {}
+resources:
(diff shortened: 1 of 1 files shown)


lefttree:
@JatinNanda I don't have azure storage access.. could you please publish the new version for me?  🙏

jeff-an:
merged before i could get to it but LGTM!

PR #82, opened 2023-02-22
[feat][DTC-3385] add liveness probe for standalone jobs runner

jeff-an:
This follows up on https://github.com/tryretool/retool_development/pull/18728 and makes the new liveness probe we developed public facing. The new probe will be by default enabled for all releases 2.110+ (the first release to contain the probe code).

The probe has been functioning in our garden environments for the past month, so we have pretty high confidence that it works.

I'm not sure how our helm charts are tested here, but helm lint didn't complain 🤷 .

### DIFF
--- a/templates/deployment_jobs.yaml
+++ b/templates/deployment_jobs.yaml
@@ -163,6 +163,15 @@ spec:
         {{- end }}
         resources:
 {{ toYaml .Values.resources | indent 10 }}
+        {{- if and (regexMatch "^[.0-9]+$" .Values.image.tag) (semverCompare ">=2.110.0" .Values.image.tag) }}
+        livenessProbe:
+          httpGet:
+            path: /api/checkJobsRunnerHealth
+            port: 3003
+          initialDelaySeconds: 120
+          periodSeconds: 30
+          timeoutSeconds: 3
+        {{- end }}
         volumeMounts:
         {{- range $configFile := (keys .Values.files) }}
         - name: {{ template "retool.name" $ }}


templates/deployment_jobs.yaml
```diff
@@ -163,6 +163,15 @@ spec:
         {{- end }}
         resources:
 {{ toYaml .Values.resources | indent 10 }}
+        {{- if and (regexMatch "^[0-9.]+$" .Values.image.tag) (semverCompare ">=2.110.0" .Values.image.tag) }}
```
ryanartecona, line 166: Do we ever release custom builds to customers, for one-off fixes? Idk if we do or not, but I was thinking of a scenario like a customer is on `2.115.0`, which would have this liveness probe enabled, and then we find some unrelated problem and give them a custom build tagged `2.115.0-finn-fix`, which would fail this regex match and get the liveness probe disabled. Thoughts?

templates/deployment_jobs.yaml
```diff
@@ -163,6 +163,15 @@ spec:
...
+        livenessProbe:
+          httpGet:
+            path: /api/checkJobsRunnerHealth
+            port: 3003
```
ryanartecona, line 170: Is this the right port to use? Asking because I don't see `3003` mentioned anywhere else in this repo.

ryanartecona (commented):
I'm not suuuper familiar with our helm charts yet so I'll leave it to Jatin to approve.

templates/deployment_jobs.yaml
jeff-an: Yeah it's the right one, it's referenced in the main retool_development repo. We're choosing a new port to avoid any conflicts that might arise in an all-in-one setup (even though we aren't adding liveness probes to that setup, just to be safe in case a customer hacks it themselves).

templates/deployment_jobs.yaml
jeff-an: Good call out, I also don't know if we do this or not to be honest. I'm hoping that if we do such patches, we don't do it via `helm`. Otherwise, I'm not sure how to handle this in a non-hacky way. I guess I could strip all alphabetical characters and then strip any trailing dashes?

templates/deployment_jobs.yaml
ryanartecona: Awesome, thanks for the details.

templates/deployment_jobs.yaml
ryanartecona: Looks like this semverCompare can already deal with these...kinda. Helm just considers them "prerelease versions" and sorts them all as _less than_ the corresponding patch version. So maybe you could just remove the `$` end anchor from your regex and that's enough? Might also be something that #deployed-engineering could chime in on.

jeff-an: thanks for investigating. removed the anchor!

JatinNanda (approved):
lg2m assuming you tested this extensively with multiple image tag value installs/upgrades and made sure `helm lint` passes

JatinNanda:
> lg2m assuming you tested this extensively with multiple image tag value installs/upgrades and made sure `helm lint` passes

here's the icky manual process for deploying a new version of our helm chart https://retoolconfluence.atlassian.net/wiki/spaces/EPD/pages/172130364/Publishing+a+new+helm+chart+version 

it's kinda easy to mess up so be careful!

PR #84, opened 2023-03-09
Replace = with : for yaml in Readme

austin-stytch:
Hi! I think the YAML only works with `:`

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Modify `values.yaml`:
             name: protos 
 
         env:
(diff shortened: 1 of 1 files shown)


avimoondra (approved):

avimoondra:
Hey @austin-stytch nice catch and thanks for the contribution! We actually moved this README into our docs - https://docs.retool.com/self-hosted/quickstarts/kubernetes/helm so will close this out.

PR #86, opened 2023-03-21
new major version including fixes for postgresql 12.1.5

carolw21:
- our current chart isn't super compatible with bitnami postgresql 12/1/5 (11.0 and higher) as there are breaking changes to how the auth data (database name, username, and password) is named and referenced (now nested under auth map with new field names)
- this changes the values template to match this new structure and also the references to it in our deployments

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 4.12.1
+version: 5.0.0
 maintainers:
   - name: Retool Engineering
     email: engineering+helm@retool.com
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -104,7 +104,7 @@ Set postgresql db
 */}}
 {{- define "retool.postgresql.db" -}}
 {{- if .Values.postgresql.enabled -}}
-{{- .Values.postgresql.postgresqlDatabase | quote -}}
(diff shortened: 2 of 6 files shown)


Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 4.12.1
+version: 5.0.0
```
JatinNanda, line 5: did we figure out why us pinning the dependent chart version didn't work? i didn't see that they got rid of the version we had pinned last i checked

carolw21: do you mean why 4.11.9 is associated with postgresql version 10.4.3 even though we pinned it to 12.1.5 in 4.11.8 release of chart? not yet, i'm working on fixing this issue so customers can work with the latest retool-helm version first. will probably download the earlier version tgz files and dissect later.

JatinNanda: oh my bad i missed the part where we pinned it to 12.1.5, makes sense

JatinNanda (approved):
seems fine as long as you've test installed it on internal onprem or something and made sure you can spin up a deployment properly

carolw21:
> seems fine as long as you've test installed it on internal onprem or something and made sure you can spin up a deployment properly

yep tested new structure and with persistent volumes and seems to work fine! just have to alter all onprem instances to get it to be compatible with the new postgresql structure 😢

PR #89, opened 2023-04-20
Implement PVC 

BenjamynBaker:
Values.persistentVolumeClaim was previously unused. This adds the ability to create or use an existing PVC and mount it to the Retool deployments. This will hopefully make it more straightforward for teams to mount volumes for protos, certs, etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 charts
+.prettierignore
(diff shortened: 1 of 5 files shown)


JatinNanda (approved):
nice

PR #98, opened 2023-05-23
Set up CI action

ryanartecona:
Add a github actions workflow to run some validations in CI.
Screen Shot 2023-05-24 at 4 11 07 PM

* `helm lint`, yaml linting, yaml schema validation on Chart.yaml via the [chart-testing](https://github.com/helm/chart-testing) tool's `ct lint` command
* K8s schema validation of the generated resources via [kubeconform](https://github.com/yannh/kubeconform)
* Test that everything comes up healthy in a local [kind](https://kind.sigs.k8s.io/) k8s cluster via `ct install`. This will work for any values files we want to add to `charts/<chartname/ci/*-values.yaml`, so we can add more of those for different deployment configurations in the future. It will basically test that containers boot up and reach a readiness probe healthy state within a timeout (5 min).

This was all heavily based on the setup in datadog's [helm-charts](https://github.com/DataDog/helm-charts) repo, which seems to work well. I swapped `kubeval` for the newer, maintained `kubeconform` fork. I also copied `ci/kubeconform-values.yaml` from our reference `values.yaml` and edited the resources and filled in blank values so it can actually boot up.

To work with the chart-testing tool, we have to move our chart files into `charts/<chartname>/`. I saw that our support docs assume in some places that the reference `values.yaml` file lives at the top level of this repo. To avoid having 2 copies that diverge, for now I moved it into `charts/retool/values.yaml` and made the top level one a symlink pointing there. My thinking is this will keep links and code snippets working for now, and we can move docs references to point at `charts/retool/values.yaml` later. Open to suggestions on how to handle that better.


JatinNanda (approved):
no idea this was even a thing, seems super useful!!

PR #101, opened 2023-05-24
set both postgres password fields in values template

mayzenator:
Both of these need to be set otherwise the postgres chart complains about the password being wrong.


carolw21 (approved):

anna-yn:
we can't add a new default because that would break the setup for anyone who installed the chart without `postgresPassword` set :(

PR #110, opened 2023-07-12
fix semver not short circuiting properly

mayzenator:
this will error if you're passing in an image tag like "latest" because helm templating isn't short circuiting this if statement


jeff-an (approved):
thanks for figuring this out!

PR #112, opened 2023-07-21
Set pod anti-affinity by default

jjlgao:
We already had this `app.kubernetes.io/name=retool` label, but it's not consistent; workflows doesn't use it. I'm going to add those labels to the workflows chart as well so that we don't need to add new labels.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -222,7 +222,18 @@ priorityClassName: ""
 
 # Affinity for pod assignment
 # Ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity
-# affinity: {}
+affinity:
+  podAntiAffinity:
+    preferredDuringSchedulingIgnoredDuringExecution:
+    - weight: 100
+      podAffinityTerm:
+        labelSelector:
+          matchExpressions:
+          - key: "app.kubernetes.io/name"
+            operator: In
+            values:
+            - retool
+        topologyKey: "kubernetes.io/hostname"
 
 # Tolerations for pod assignment
 # Ref: https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/
--- a/charts/retool/Chart.yaml
+++ b/charts/retool/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 5.0.8
+version: 5.0.9
 maintainers:
   - name: Retool Engineering
     email: engineering+helm@retool.com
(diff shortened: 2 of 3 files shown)


values.yaml
```diff
@@ -222,7 +222,18 @@ priorityClassName: ""
...
+    preferredDuringSchedulingIgnoredDuringExecution:
+    - weight: 100
+      podAffinityTerm:
+        labelSelector:
+          matchExpressions:
+          - key: "app.kubernetes.io/name"
+            operator: In
+            values:
+            - retool
+        topologyKey: "kubernetes.io/hostname"
```
june-xu, line 237: what's the differentiation between the two values files?

june-xu: cc @jjlgao

jjlgao: i have no idea lol, i think @ryanartecona moved it to use only the inner one but I wanted to modify both to keep them the same.

june-xu: ahh I see -- if the outer one is not being used, wonder if we should just remove it? we don't have to do it on this PR though

june-xu (approved):
thanks for adding these changes!

values.yaml
jjlgao: I agree -- it's confusing, but I personally don't know for sure.

jjlgao:
Tested, the affinity functionality appears to work, so I'm going to merge and release!

values.yaml
ryanartecona: ah yeah. I had to move all the chart files into charts/retool/ because the linter and validator tools I setup needed them to be there, but I also needed to leave a top level values.yaml file where it was because lots of our docs reference them, and supposedly some customers have forks or scripts that expect that to stay there. and also a symlink didn't work because it broke something else, I don't remember. so the best I could do was to duplicate (😭) them and add a check that makes sure they always stay in sync.

june-xu: np, ty for the context! maybe we should add this context as a comment on top of the values file? just to make sure no one accidentally deletes it / removes it

ryanartecona: we could for extra context if that helps, but there's definitely already a CI check that will fail if these yaml files diverge at all [here](https://github.com/tryretool/retool-helm/blob/main/.github/workflows/values-yaml-synced.yaml) (you can see it in the Checks tab on this PR too). that will fail if someone edits one but not the other, or if one gets deleted, both.

jjlgao: It won't help for external folks who have to fork the repo and create a PR cross-repo. I noticed those don't kick off PR checks.

june-xu: ahh true true! all good here then!

ryanartecona: ahhh right you are. unfortunate. I think those have to be approved explicitly by a repo maintainer? looking at recent PR history we've only had like 2 or 3 external contributions in the last 2 months, so hopefully that's low enough volume to just give a bit more guidance to those.

PR #114, opened 2023-08-04
use the correct password for the postgresql subchart based on username

anna-yn:
Right now, by default, we have these values:
```
postgresql:
  # We highly recommend you do NOT use this subchart as is to run Postgres in a container
  # for your production instance of Retool; it is a default. Please use a managed Postgres,
  # or self-host more permanantly. Use enabled: false and set in config above to do so.
  enabled: true
  ssl_enabled: false
  auth:
    database: hammerhead_production
    username: postgres
    password: retool
```
`postgres` happens to be the username of the default admin user in the postgresql subchart, which means it needs the admin password `postgresPassword`. These defaults would work out of the box because retool will be initiated with the user `postgres` and the auto-generated admin password `postgresPassword` which we did not specify here.

however, if the user changes username to anything else other than `postgres`, the install would break unless the user sets a `postgresPassword` field that has the same value as the password field. This is because we only pass the value of `postgresPassword` to retool regardless of whether password is set, but by specifying a custom user that is not `postgres`, the pg subchart actually expects `password` to be used for the custom user, not `postgresPassword`.

This fixes this case by checking whether username == "postgres", if true, we’d use `postgresPassword` , otherwise we use `password`. Also added some comments to warn people about this situation

We can't just add a new default for `postgresPassword` because that would break anyone who has installed the chart with the username `postgres` but not have `postgresPassword` set by changing the password for them.

I tested:
1. pg username postgres , postgresPassword is used (pic1):
  a. new installation works
  b. upgrading to this new chart version doesn’t change anything
2. pg username retool , password is used (pic2)
  a. new installation works
  b. upgrading to this new chart version changes the secret key from postgres-password to password which doesn’t break anything - if anyone has username set to anything other than postgres right now, the only way they could’ve made it work is to set `postgresPassword to be the same value as password, so changing the secret reference doesnt change anything

pic1:
image
pic2:
image

### DIFF
--- a/charts/retool/values.yaml
+++ b/charts/retool/values.yaml
@@ -154,6 +154,9 @@ postgresql:
     database: hammerhead_production
     username: postgres
     password: retool
+    # Uncomment this if setting `username` to `postgres`. If this field is not set, a random password
+    # will be generated and used. And you might run into password mismatch issues when re-installing the chart.
+    # postgresPassword: retool
   service:
     port: 5432
   # Use the offical docker image rather than bitnami/docker
--- a/charts/retool/values.yaml
+++ b/charts/retool/values.yaml
@@ -154,8 +154,7 @@ postgresql:
     database: hammerhead_production
     username: postgres
     password: retool
-    # Uncomment this if setting `username` to `postgres`. If this field is not set, a random password
-    # will be generated and used. And you might run into password mismatch issues when re-installing the chart.
+    # IMPORTANT: When setting username to `postgres` a random password will be generated unless the postgresPassword field is uncommented
     # postgresPassword: retool
   service:
     port: 5432
--- a/values.yaml
+++ b/values.yaml
@@ -154,6 +154,9 @@ postgresql:
     database: hammerhead_production
     username: postgres
     password: retool
+    # Uncomment this if setting `username` to `postgres`. If this field is not set, a random password
+    # will be generated and used. And you might run into password mismatch issues when re-installing the chart.
+    # postgresPassword: retool
   service:
     port: 5432
   # Use the offical docker image rather than bitnami/docker
--- a/values.yaml
+++ b/values.yaml
@@ -154,8 +154,7 @@ postgresql:
     database: hammerhead_production
     username: postgres
     password: retool
-    # Uncomment this if setting `username` to `postgres`. If this field is not set, a random password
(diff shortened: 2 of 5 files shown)


charts/retool/values.yaml
```diff
@@ -154,6 +154,9 @@ postgresql:
     database: hammerhead_production
     username: postgres
     password: retool
+    # Uncomment this if setting `username` to `postgres`. If this field is not set, a random password
```
JatinNanda, line 157: why is this by default commented out if the username is by default set to postgres?

values.yaml
```diff
@@ -154,6 +154,9 @@ postgresql:
     database: hammerhead_production
     username: postgres
     password: retool
+    # Uncomment this if setting `username` to `postgres`. If this field is not set, a random password
```
JatinNanda, line 157: ditto

charts/retool/values.yaml
anna-yn: We can't just add a new default for postgresPassword because that would break anyone who has installed the chart with the username postgres but not have postgresPassword set by changing the password for them.

same thing with the username, we can't just change the default now that it's already been `postgres` for forever. there might be people using these default values and not specifying their own at all, and if we changed it upgrading would break things for them. the best thing we can do here is to make sure the upgrades and future new installs dont break

JatinNanda: does it make sense to comment out the username field too then? I feel like as is the wording of the comment makes me wonder if i should uncomment it

like basically force the user to end up having to either uncomment (hopefully im following this right)
#username: postgres
#postgresPassword: defaultpassword

or 
#username: your-custom-user
#password: defaultpassword

anna-yn: that's still changing the defaults though. a user can have a values.yaml file that only has the license key, jwt secret and encryption key set, and that would give them a working retool instance using the `postgres` user + generated password combo right now. if we comment that out from our defaults, their next upgrade will break

JatinNanda: oh oops ok

JatinNanda: can we change the wording to 'IMPORTANT: When setting `username` to `postgres` a random password will be generated unless this field is uncommented'

I think my issue is with the current wording being 'if ur setting something, do x' but it already appears to be set by default

JatinNanda (approved):
ty for untangling this mess :')

Issue #57, opened 2022-06-05
Chart does not respect securityContext settings


jawnsy:
# Current behavior

The chart looks for a `securityContext.enabled` setting, which sets a default security context:

https://github.com/tryretool/retool-helm/blob/1e06cee2744e41197b66eb501c467cddd253c783/templates/deployment_backend.yaml#L219-L223

However, the default values file implies that other securityContext settings can be set on the pod, but in fact, those values are ignored:

https://github.com/tryretool/retool-helm/blob/1e06cee2744e41197b66eb501c467cddd253c783/values.yaml#L271-L276

# Desired behavior

We should be able to configure all of the settings defined in the PodSecurityContext API: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#podsecuritycontext-v1-core

```yaml
securityContext:
  fsGroup: 1000
  fsGroupChangePolicy: OnRootMismatch
  runAsGroup: 1000
  runAsNonRoot: true
  runAsUser: 1000
```

I would also expect that we can configure all the [SecurityContext settings](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#securitycontext-v1-core) for the primary retool container

```yaml
securityContext:
  allowPrivilegeEscalation: false
  privileged: false
  runAsNonRoot: true
  runAsUser: 1000
  runAsGroup: 1000
  readOnlyRootFilesystem: true
```

These customizations are required because some environments, such as OpenShift, or clusters that enforce restrictive policies using tools like Open Policy Agent Gatekeeper or Kyverno, will not be able to install retool otherwise.

# Steps to reproduce

You can see this by running a command like:

```shell-session
$ helm template retool retool/retool \
  --set=image.tag=latest \
  --set=securityContext.enabled=true \
  --set=postgresql.enabled=false \
  --set=ingress.enabled=false \
  --set=securityContext.runAsNonRoot=true
```

You will see output like the following:

```yaml
      securityContext:
        runAsUser: 1000
        fsGroup: 2000
```

It's not possible to override the `runAsUser`, to add `runAsGroup` information, or to set `runAsNonRoot: true`

# Other information

Here's my chart version information:

```shell-session
$ helm search repo retool
NAME            CHART VERSION   APP VERSION     DESCRIPTION                
retool/retool   4.10.0                          A Helm chart for Kubernetes
```

jjlgao:
This should be resolved in chart versions 6.3.2 and beyond.

PR #117, opened 2023-08-30
Add allowPrivilegeEscalation to the securityContexts

beahues:
It's mentioned in `values.yaml` but not actually reflected in the code,
but enable configuring the PodSecurity policy option for
`allowPrivilegeEscalation` to the chart.


beahues:
Addresses a tiny amount of #57

cedenilla:
can we include `runAsNonRoot` as well? It will help us a lot and it's a minimal change.

lmb-djaquier:
Hi!

Could be also great to add readOnlyRootFilesystem: true

And merge that PR as well :-) thanks for the job!

jjlgao:
Will supersede this PR with this change, as our helm chart has migrated quite a bit: https://github.com/tryretool/retool-helm/pull/206

PR #121, opened 2023-09-26
allow setting of envvars for temporal deployment

roberto-retool:


charts/retool/charts/retool-temporal-services-helm/templates/server-deployment.yaml
```diff
@@ -184,6 +184,9 @@ spec:
             - name: TEMPORAL_VERSION_CHECK_DISABLED
               value: "1"
           {{- end }}
+          {{- with .Values.environmentVariables }}
+{{ toYaml . | indent 10 }}
```
avimoondra, line 188: 10 or 12? 

can template this to make sure with `helm template -f ~/retool-helm/charts/retool/values.yaml --set config.encryptionKey="foo"  foo ~/retool-helm/charts/retool` (replacing with appropriate chart and paths etc.)

avimoondra (approved):

charts/retool/charts/retool-temporal-services-helm/templates/server-deployment.yaml
roberto-retool: yup hit this when testing; great catch!

PR #122, opened 2023-10-13
[feat][DTC-5411] updated helm chart for jobs runner temporal access

Sthakur27:
Updates jobs runner to put in the workflows env vars if workflows/temporal is enabled
https://linear.app/retool/issue/DTC-5411/update-docker-images-with-new-jobs-runner-env-vars

### DIFF
--- a/charts/retool/Chart.yaml
+++ b/charts/retool/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 6.0.3
+version: 6.0.4
 maintainers:
   - name: Retool Engineering
     email: engineering+helm@retool.com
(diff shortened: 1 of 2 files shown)


avimoondra (approved):
LGTM

- if you are on a version of retool that cannot leverage temporal, setting these does not affect anything.
- if you are on a version of retool that can leverage temporal -- if temporal is not configured, jobs runner continues as is; if temporal is configured, jobs runner starts as internal worker

charts/retool/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 6.0.3
+version: 6.0.4
```
avimoondra, line 5: i think 6.0.5 now

avimoondra (commented):
When we merge, just a reminder to ourselves that in the release notes (and in docs as needed) to make sure we identify what retool version is needed to support this change.

PR #130, opened 2023-10-26
Add readinessProb to jobs container

flatiron32:
My platform team has implemented an admission controller that requires a readinessProbe to be defined. Using the livelinessProbe as a model, this implements it for the jobs container.


avimoondra (approved):

avimoondra:
Thanks for the PR @flatiron32 !

PR #132, opened 2023-11-20
Support granular UI annotations and labels

woz5999:
This PR adds the ability to specifically target the UI pods with annotations. The current available annotations configurations apply to multiple deployments, including the UI pods. In order to expose the service via our service mesh, we need the ability to target the UI pods with annotations. This change allows us to inject and expose the UI pods only, without inadvertently annotated undesired pods. 

if you need me to make any changes like bumping versions or updating changelogs, etc, just let me know and I'm happy to do so.

### DIFF
--- a/charts/retool/Chart.yaml
+++ b/charts/retool/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 6.0.6
+version: 6.0.9
 maintainers:
   - name: Retool Engineering
     email: engineering+helm@retool.com
--- a/charts/retool/templates/deployment_backend.yaml
+++ b/charts/retool/templates/deployment_backend.yaml
@@ -23,6 +23,9 @@ spec:
 {{- if .Values.backend.annotations }}
 {{ toYaml .Values.backend.annotations | indent 8 }}
 {{- end }}
+{{- if .Values.ui.annotations }}
+{{ toYaml .Values.ui.annotations | indent 8 }}
+{{- end }}
       labels:
 {{- include "retool.selectorLabels" . | nindent 8 }}
 {{- if .Values.podLabels }}
@@ -30,6 +33,9 @@ spec:
 {{- end }}
 {{- if .Values.backend.labels }}
 {{ toYaml .Values.backend.labels | indent 8 }}
+{{- end }}
+{{- if .Values.ui.labels }}
+{{ toYaml .Values.ui.labels | indent 8 }}
 {{- end }}
     spec:
       serviceAccountName: {{ template "retool.serviceAccountName" . }}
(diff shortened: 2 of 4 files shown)


bdjohnson529:
@woz5999 Are the UI pods the pods within the [backend deployment](https://github.com/tryretool/retool-helm/blob/main/charts/retool/templates/deployment_backend.yaml)? [Those pods would be](https://github.com/tryretool/retool-helm/blob/main/charts/retool/templates/deployment_backend.yaml#L68):

- `MAIN_BACKEND`
- `DB_CONNECTOR`
- `DB_SSH_CONNECTOR`
- `JOBS_RUNNER`

I can see that the `backend` annotation is shared with the [workflows](https://github.com/tryretool/retool-helm/blob/main/charts/retool/templates/deployment_workflows.yaml#L25) and [workflows worker](https://github.com/tryretool/retool-helm/blob/main/charts/retool/templates/deployment_workflows_worker.yaml#L28) deployments.

Is your primary concern having a separate annotation which exclusively targets pods within the `backend` deployment?

woz5999:
@bdjohnson529 That's correct. Our service mesh leverages annotations for registration and routing. Currently, if we set the annotation for the "backend" is applies to the workflow pods as well, meaning that UI requests regularly get round robin'ed to non-UI pods, resulting in HTTP errors. By being able to target the UI pods individually, the issue is totally resolved.

jjlgao:
Hey @woz5999 , Jason from Retool's infra team -- the PR makes sense to me, but we don't internally call these non-workflows pods "ui", so checking with our team on what we'd like that naming to be.

woz5999:
Sure thing. Naming's hard. Just let me know when you're ready and I can happily update.

jjlgao:
Hey @woz5999 , we talked internally and we're good with using `ui`, but I just realized that we have a requirement for the two values.yaml files to match in the repo for legacy reasons. There's this topmost values.yaml file if you could copy your change over as well: https://github.com/tryretool/retool-helm/blob/main/values.yaml

Once you do that I'll merge and release this change for you.

Thanks!

woz5999:
@jjlgao done. let me know if you need anything else

woz5999:
The linter is failing because the chart version needs to be updated. I'm happy to do that. 
``` ✖︎ retool => (version: "6.0.8", path: "charts/retool") > chart version not ok. Needs a version bump! ```

I'm guessing 6.1.0 should be the new version?

jjlgao:
> The linter is failing because the chart version needs to be updated. I'm happy to do that. `✖︎ retool => (version: "6.0.8", path: "charts/retool") > chart version not ok. Needs a version bump!`
> 
> I'm guessing 6.1.0 should be the new version?

6.0.9, thanks!

woz5999:
Done

woz5999:
Thanks everyone. When do new releases get published to your helm repository?

jjlgao:
They're done manually. I'm going to publish it in a bit.

jjlgao:
Just published, try pulling now!

PR #143, opened 2024-01-25
allow name overrides in pods

avimoondra:
Enable internal Retool to align service names with Cloud Prod, for DD dashboards shared across env

----- 

## before the change

without custom names:

```
helm template -f ~/retool-helm/charts/retool/values.yaml retool ~/retool-helm/charts/retool --set config.encryptionKey="foo" --set image.tag="5.6.10" --set codeExecutor.image.tag="5.6.10" --set codeExecutor.enabled=true | rg retool- -C 2
```

```
kind: Secret
metadata:
  name: retool-postgresql
  namespace: "default"
  labels:
--
  name: retool
  labels:
    helm.sh/chart: retool-6.0.13
    app.kubernetes.io/name: retool
    app.kubernetes.io/instance: retool
--
kind: Service
metadata:
  name: retool-postgresql-hl
  namespace: "default"
  labels:
--
kind: Service
metadata:
  name: retool-postgresql
  namespace: "default"
  labels:
--
kind: Service
metadata:
  name: retool-code-executor
spec:
  selector:
    retoolService: retool-code-executor
  ports:
  - protocol: TCP
--
kind: Service
metadata:
  name: retool-workflow-backend
spec:
  selector:
    retoolService: retool-workflow-backend
  ports:
  - protocol: TCP
--
kind: Service
metadata:
  name: retool-workflow-worker
spec:
  selector:
    retoolService: retool-workflow-worker
  ports:
  - protocol: TCP
--
metadata:
  labels:
    helm.sh/chart: retool-6.0.13
    app.kubernetes.io/name: retool
    app.kubernetes.io/instance: retool
--
  name: retool
  labels:
    helm.sh/chart: retool-6.0.13
    app.kubernetes.io/name: retool
    app.kubernetes.io/instance: retool
--
          - bash
          - -c
          - chmod -R +x ./docker_scripts; sync; ./docker_scripts/wait-for-it.sh -t 0 "retool-postgresql":"5432"; ./docker_scripts/start_api.sh
        env:
          - name: NODE_ENV
--
            value: "false"
          - name: POSTGRES_HOST
            value: "retool-postgresql"
          - name: POSTGRES_PORT
            value: "5432"
--
            value: "false"
          - name: WORKFLOW_BACKEND_HOST
            value: http://retool-workflow-backend
          - name: CODE_EXECUTOR_INGRESS_DOMAIN
            value: http://retool-code-executor
          - name: LICENSE_KEY
            valueFrom:
--
            valueFrom:
[...]

PR #146, opened 2024-02-02
Revert "allow name overrides in pods"

avimoondra:
Reverts tryretool/retool-helm#143

### DIFF
--- a/charts/retool/Chart.yaml
+++ b/charts/retool/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 6.0.14
+version: 6.0.13
 maintainers:
   - name: Retool Engineering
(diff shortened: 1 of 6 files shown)


charts/retool/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 6.0.14
+version: 6.0.13
```
jjlgao, line 5: i think you can only up this to 6.0.15

avimoondra: I think I can go back to 6.0.13 if I removed the release from Azure, no? 

https://retoolconfluence.atlassian.net/wiki/spaces/EPD/pages/172130364/Publishing+a+new+helm+chart+version

jjlgao (approved):

PR #153, opened 2024-02-23
Add nodeSelector to code-executor as well

erulabs:
Because retool images are not built for ARM64, a nodeSelector `kubernetes.io/arch: amd64` needs to be added to all mixed architecture clusters.

Note that a reasonable change, in my mind, is to make `kubernetes.io/arch: amd64` a default node selector. That should work on almost all clusters and will prevent crashes on mixed architecture clusters.


avimoondra:
Hey @erulabs  this is addressed in https://github.com/tryretool/retool-helm/pull/169. Thanks for the PR!

PR #154, opened 2024-02-28
add configurable port names for istio compatibility

avimoondra:


charts/retool/templates/deployment_workflows_worker.yaml
```diff
@@ -336,9 +340,13 @@ spec:
   - protocol: TCP
     port: 3005
     targetPort: 3005
+    {{ if .Values.workflows.worker.portName }}
+    name: {{ .Values.workflows.worker.portName }}
+    {{ else }}
     name: {{ template "retool.name" . }}
```
alexlo03, line 346: Maybe you can do the `| default` here too

https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function

alexlo03:
LGTM

1. you need to update values.yaml OR document these possible values
2. IMO there should be a default port name all the time, I believe that is a non-breaking change

charts/retool/templates/deployment_code_executor.yaml
```diff
@@ -137,9 +138,9 @@ spec:
...
     port: 80
     targetPort: 3004
+    {{ if .Values.codeExecutor.portName }}
+    name: {{ .Values.codeExecutor.portName }}
+    {{ else }}
     name: {{ template "retool.name" . }}
-  - protocol: TCP
-    port: 9090
-    targetPort: metrics
-    name: metrics
```
avimoondra, line 144: unused, no need to expose this.

charts/retool/templates/deployment_workflows_worker.yaml
avimoondra: line 346 is the fallback if `.Values.workflows.worker.portName` is not set, so we should be good here?

alexlo03: I mean like you do above - the conditional L343-L347 becomes one line

`name: {{ .Values.workflows.worker.portName | default (template "retool.name" .) }}`

charts/retool/templates/deployment_code_executor.yaml
```diff
@@ -137,9 +138,9 @@ spec:
...
+    {{ end }}
```
alexlo03, line 145: can use | default here too

charts/retool/templates/deployment_code_executor.yaml
```diff
@@ -83,10 +83,11 @@ spec:
...
         ports:
         - containerPort: 3004
+          {{ if .Values.codeExecutor.portName }}
+          name: {{ .Values.codeExecutor.portName }}
+          {{ else }}
           name: {{ template "retool.name" . }}
-          protocol: TCP
-        - containerPort: 9090
-          name: metrics
+          {{ end }}
```
alexlo03, line 90: and here

alexlo03:
LGTM, the defaulting comments are not necessary but I think you could collapse the conditionals down

avimoondra:
> LGTM, the defaulting comments are not necessary but I think you could collapse the conditionals down

The if/else is giving that default - but I ran into an issue using default with "template"... e.g. this wasn't working: 

{{ .Values.codeExecutor.portName | default (template "retool.name" .) }}

neither this

{{ default  .Values.codeExecutor.portName (template "retool.name" .) }}

lmk if you have a different suggestion @alexlo03 ! thanks for review.

charts/retool/values.yaml
```diff
@@ -329,12 +329,15 @@ workflows:
     # Scaling this number will increase the number of workflow workers, e.g. a replicaCount of 4
     # will launch 9 pods -- 1 workflow backend, 4 workflow workers, and 4 for temporal cluster
     replicaCount: 1
+    # portName: custom-port-name, e.g. http-workflows-worker; defaults to "retool.name"
+    # metricsPortName: custom-port-name, e.g. http-workflow-worker-metrics; defaults to "metrics"
```
ryanartecona, line 333: For all the defaults in our values.yaml, I think it would make sense to just pick a default port name and possibly just drop the `template "retool.name" .` and `| default "metrics"` dynamic fallbacks. The defaults can be static (i.e. no point to using `template "retool.name"` anyway) and we can even just use the istio naming convention for them. This would remove the need for those conditionals in templates and just simplify a bit. Adding port names is a non-breaking change anyway I am pretty sure, so extremely low risk.

ryanartecona (commented):
I haven't looked through the whole chart in this branch, but do we want to use these new port names for the corresponding Service resources? i.e. https://kubernetes.io/docs/concepts/services-networking/service/#field-spec-ports. I believe a Service can be associated with a port either via port number or port name, so we don't have to, but we'd have the option now with this change. Not a big deal to skip tho.

ryanartecona:
> I ran into an issue using default with "template"...

oh I didn't notice this comment before, but that's because `template` doesn't return the value, it just prints it out to the output. you can just change `template` to `include`, which returns instead of printing, and then it will work as expected with the default pipeline.

avimoondra:
> > I ran into an issue using default with "template"...
> 
> oh I didn't notice this comment before, but that's because `template` doesn't return the value, it just prints it out to the output. you can just change `template` to `include`, which returns instead of printing, and then it will work as expected with the default pipeline. usually include is preferred over template in helm for exactly this reason.

Ah thanks @ryanartecona - will address all of it w/your feedback today! 🙏

alexlo03:
gentle ping. would like to see this in before we upgrade charts. thanks 🙏

avimoondra:
Closing in favor of https://github.com/tryretool/retool-helm/pull/170

PR #156, opened 2024-03-04
Remove prom scrape annotations from code-executor, which does not have metrics

### LINKED ISSUE #155: Code executor does not expose metrics endpoint but does have metrics annotations
The `retool-code-executor` deployment [contains the prometheus scrape annotations](https://github.com/tryretool/retool-helm/blob/main/charts/retool/templates/deployment_code_executor.yaml#L23-L24):

        prometheus.io/scrape: 'true'
        prometheus.io/port: '9090'

However, there is no process listening on 9090 in code executor:

```
root@retool-code-executor-8578784858-w9m2f:/retool/code_executor# curl localhost:9090
curl: (7) Failed to connect to localhost port 9090: Connection refused
root@retool-code-executor-8578784858-w9m2f:/retool/code_executor# netstat -pnlt
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name
tcp6       0      0 :::3004                 :::*                    LISTEN      14/node
```

We should remove the prometheus scrape annotations from code executor.

erulabs:
This prevents error-spam from failures to scrape metrics.

Resolves #155

### DIFF
--- a/charts/retool/templates/deployment_code_executor.yaml
+++ b/charts/retool/templates/deployment_code_executor.yaml
@@ -19,9 +19,6 @@ spec:
   template:
     metadata:
       annotations:
-        prometheus.io/job: {{ template "retool.codeExecutor.name" . }}
-        prometheus.io/scrape: 'true'
-        prometheus.io/port: '9090'
 {{- if .Values.podAnnotations }}
 {{ toYaml .Values.podAnnotations | indent 8 }}
 {{- end }}


samsternatretool:
@erulabs thank you for submitting this PR! Let me find the right person to review it.

ryanartecona:
ah yes, good find. this happened to be noticed independently and will be fixed (in the same way as here) in #154, which should be the next release to go out, soon. cc @avimoondra

avimoondra:
> ah yes, good find. this happened to be noticed independently and will be fixed (in the same way as here) in #154, which should be the next release to go out, soon. cc @avimoondra

🎯  should be resolved in a day or two!

samsternatretool:
Sounds like we can close this one, thanks @erulabs for the contribution though!

PR #157, opened 2024-03-07
Add `WORKFLOW_BACKEND_HOST` to code-executor deployment

tomknig:
If retool workflows are enabled, `code-executor` deployments need to know which workflow backend to talk to. Before this update, users of the chart would have to inject this env var manually.

With this PR, I am proposing to add the (conditional) environment variable `WORKFLOW_BACKEND_HOST` to the `code-executor` deployment in the same way that it is being added in the other deployment manifests.


avimoondra:
Hey @tomknig thanks for the PR. The code executor receives the workflow backend host dynamically for each request, so there is not a fixed workflow backend host that can be respected in environment variables. Will go ahead and close out this PR!

PR #158, opened 2024-03-07
Add telemetry support

ryanartecona:
This introduces a `telemetry` deployment and supporting resources which, when enabled, will collect metrics and logs from the other pods in the chart. When the telemetry deployment is enabled, those metrics and logs will be sent to Retool, for use cases such as proactively finding retool bugs which manifest in self-hosted deployment logs and having an easier path for Retool staff to diagnose deployment issues in a self-hosted deployment.


charts/retool/templates/_helpers.tpl
```diff
@@ -52,13 +52,67 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
...
 app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
 
+{{/*
+Selector labels for workflow backend. Note changes here will require manual
+deployment recreation and incur downtime, so should be avoided.
+*/}}
+{{- define "retool.workflowBackend.selectorLabels" -}}
+retoolService: {{ template "retool.workflowBackend.name" . }}
+{{- end }}
```
ryanartecona, line 70: This and all the `retool.*.selectorLabels` partials below don't actually change the selector labels that were already used, they're just refactoring the template code.

charts/retool/templates/_helpers.tpl
```diff
@@ -230,6 +284,20 @@ Set Temporal namespace
...
+{{- define "retool.workflowBackend.name" -}}
+{{ template "retool.fullname" . }}-workflow-backend
+{{- end -}}
+
+{{/*
+Set workflow worker service name
+*/}}
+{{- define "retool.workflowWorker.name" -}}
+{{ template "retool.fullname" . }}-workflow-worker
+{{- end -}}
```
ryanartecona, line 299: These result in the exact same names as before, just a template code refactor for reuse.

charts/retool/templates/deployment_backend.yaml
```diff
@@ -54,7 +55,7 @@ spec:
 {{- end }}
 {{- end }}
       containers:
-      - name: {{ .Chart.Name }}
+      - name: main-backend
```
ryanartecona, line 58: container names are mutable, since the pods just get recreated as normal. using semantic container names helps log collection and metric collection, since deployment name isn't always available and pod name is dependent on nameOverride and such.

charts/retool/templates/deployment_backend.yaml
```diff
@@ -244,6 +248,7 @@ spec:
             port: {{ .Values.service.internalPort }}
           initialDelaySeconds: {{ .Values.livenessProbe.initialDelaySeconds }}
           timeoutSeconds: {{ .Values.livenessProbe.timeoutSeconds }}
+          periodSeconds: {{ .Values.livenessProbe.periodSeconds }}
```
ryanartecona, line 251: this was present in values.yaml but not in the template, so I assume just an oversight.

charts/retool/templates/telemetry_deployment.yaml
```diff
@@ -0,0 +1,158 @@
...
+    matchLabels:
+      {{- include "retool.telemetry.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        {{- include "retool.telemetry.selectorLabels" . | nindent 8 }}
+      annotations:
+        {{- if or .Values.telemetry.customVectorConfig .Values.telemetry.customGrafanaAgentConfig }}
+        checksum/custom-config: {{ include (print $.Template.BasePath "/telemetry_configmap.yaml") . | sha256sum | quote }}
+        {{- end }}
```
ryanartecona, line 19: this makes it so that custom config changes will still roll the telemetry pod even if the pod itself doesn't change

KnlnKS (approved):

charts/retool/templates/telemetry_deployment.yaml
avimoondra, line 1: `deployment_telemetry.yaml` perhaps for consistency

charts/retool/templates/telemetry_deployment.yaml
```diff
@@ -0,0 +1,158 @@
+{{- if .Values.telemetry.enabled }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "retool.fullname" . }}-telemetry
+  labels:
+    {{- include "retool.telemetry.labels" . | nindent 4 }}
```
avimoondra, line 7:
```suggestion
    {{- include "retool.telemetry.labels" . | nindent 4 }}
    {{- include "retool.telemetry.selectorLabels" . | nindent 4 }}
```

charts/retool/templates/telemetry_deployment.yaml
```diff
@@ -0,0 +1,158 @@
...
+          hostPath:
+            path: "/sys"
+
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "retool.telemetry.fullname" . }}
+  labels:
+    {{- include "retool.telemetry.labels" . | nindent 4 }}
```
avimoondra, line 147:
```suggestion
    {{- include "retool.telemetry.labels" . | nindent 4 }}
    {{- include "retool.telemetry.selectorLabels" . | nindent 4 }}
```

charts/retool/templates/telemetry_deployment.yaml
```diff
@@ -0,0 +1,158 @@
...
+                key: {{ .Values.config.licenseKeySecretKey | default "license-key" }}
+                {{- else }}
+                name: {{ template "retool.fullname" . }}
+                key: license-key
+                {{- end }}
+          {{- end }}
+          - name: RTEL_DEPLOYMENT_MODE
+            value: 'kubernetes-helm'
+          - name: RTEL_HELM_RELEASE_NAME
+            value: {{ .Release.Name | quote }}
```
avimoondra, line 61: maybe: align this to https://github.com/tryretool/retool_development/blob/dev/backend/src/server/modules/jobsRunner/startupTelemetry.ts#L49-L62

DEPLOYMENT_TEMPLATE_TYPE
DEPLOYMENT_TEMPLATE_VERSION

charts/retool/templates/_helpers.tpl
```diff
@@ -44,21 +44,74 @@ Common labels
 */}}
 {{- define "retool.labels" -}}
 helm.sh/chart: {{ include "retool.chart" . }}
-{{ include "retool.selectorLabels" . }}
 {{- if .Chart.AppVersion }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
```
avimoondra, line 48: what is AppVersion vs. .Chart.Version ?  (only diff from retool.telemetry.labels)

I see it's used by retool-temporal-services-helm only 👀

avimoondra (approved):
🚢

charts/retool/templates/telemetry_deployment.yaml
avimoondra: not sure if needed, but for consistency as well

charts/retool/templates/deployment_backend.yaml
avimoondra: let's go with "main-backend" even tho this has services bundled into it (including db connector)

"api" is not as discoverable in docs.

charts/retool/templates/telemetry_deployment.yaml
ryanartecona: I think I will need to keep the env vars `RTEL_DEPLOYMENT_MODE` and `DEPLOYMENT_TEMPLATE_TYPE` separate after all, because I think they will diverge for other deployment templates. I'll still at least unify the values so e.g. `k8s-helm` is used by both, and also add `DEPLOYMENT_TEMPLATE_TYPE` as tags to metrics/logs when present.

charts/retool/templates/_helpers.tpl
ryanartecona: pretty sure `.Chart.Version` is the chart version, and `.Chart.AppVersion` would be the retool version. except `.Chart.AppVersion` would come from `appVersion:` in Chart.yaml, which is optional and absent for our chart. so I think `.Chart.AppVersion` is effectively always null, and we don't use it to specify retool version anyway. docs: https://helm.sh/docs/topics/charts/

I didn't realize all this when I first wrote this here though, so I'm just going to take it out so it doesn't confuse us further.

charts/retool/templates/telemetry_deployment.yaml
avimoondra: ack ok 👍

PR #163, opened 2024-03-22
set user for visibility db in VISIBILITY_DBUSER env var

avimoondra:
Support using different users, each scoped to the DB (default store, visibility store) they are expected to manage. in our auto set up scripts (https://github.com/tryretool/retool-temporal-docker-builds/pull/2) we assume the same user is used to set up schema for both these databases.

see issue here: https://retooled.slack.com/archives/C05TR7RRB0F/p1710327058225469

relates to https://github.com/tryretool/retool-temporal-docker-builds/pull/2

### DIFF
--- a/charts/retool/charts/retool-temporal-services-helm/values.yaml
+++ b/charts/retool/charts/retool-temporal-services-helm/values.yaml
@@ -25,7 +25,7 @@ server:
   autosetup: true
   image:
     repository: tryretool/one-offs
-    tag: retool-temporal-1.1.3
+    tag: retool-temporal-1.1.5
     pullPolicy: IfNotPresent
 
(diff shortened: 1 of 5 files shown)


charts/retool/charts/retool-temporal-services-helm/values.yaml
```diff
@@ -25,7 +25,7 @@ server:
   autosetup: true
   image:
     repository: tryretool/one-offs
-    tag: retool-temporal-1.1.3
+    tag: retool-temporal-1.1.5
```
avimoondra, line 28: the main change, rest is formatting

demitrin (approved):

PR #166, opened 2024-03-28
Update values.yaml

avimoondra:


avimoondra:
@BenjamynBaker  this doesn't need a version bump if you have admin privs 🙏

BenjamynBaker (approved):

BenjamynBaker:
I think you may need to update chart/retool/values.yaml as well

PR #169, opened 2024-05-17
fix: add a missing `nodeSelector` field to the code-executor deployment

criskurtin:
This PR addresses the issue of a missing `nodeSelector` in the `code-executor` deployment, which causes problems when the cluster is based on mixed-architecture nodes. For example, in our case, the `code-executor` pod was provisioned on an `arm64` node.

### DIFF
--- a/charts/retool/templates/deployment_code_executor.yaml
+++ b/charts/retool/templates/deployment_code_executor.yaml
@@ -131,6 +131,10 @@ spec:
       affinity:
 {{ toYaml .Values.codeExecutor.affinity | indent 8 }}
 {{- end }}
+{{- if .Values.nodeSelector }}
+      nodeSelector:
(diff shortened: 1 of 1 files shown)


avimoondra (approved):

chawleejay:
has anyone run into any restart issues when adding node selector fields to your deployment? trying to run retool on its own nodepool, and when I add tolerations to the code-executor and other backend pieces the pods spin up successfully then give me this warning & constantly restart.


`(node:50) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
(Use `node --trace-deprecation ...` to show where the warning was created)`

Issue #144, opened 2024-01-31
Support Istio Clusters


alexlo03:
Hello,

We are able to use version: 5.0.10 in our Istio cluster, but not 6.0.13 because the chart renders things that do not pass our Istio checks.

```
Error: 3 errors occurred:
	* Service/my-retool/my-retool-workflow-backend: service "my-retool-workflow-backend/my-retool/:" has an unnamed port. This is not recommended, See https://istio.io/v1.19/docs/ops/deployment/requirements/
	* Service/my-retool/my-retool-workflow-worker: service "my-retool-workflow-worker/my-retool/:" port "retool" does not follow the Istio naming convention. See https://istio.io/v1.19/docs/ops/deployment/requirements/
	* Service/my-retool/my-retool-workflow-worker: service "my-retool-workflow-worker/my-retool/:" port "metrics" does not follow the Istio naming convention. See https://istio.io/v1.19/docs/ops/deployment/requirements/
```

# Error 1

port is not named - issue here: https://github.com/tryretool/retool-helm/blob/main/charts/retool/templates/deployment_workflows.yaml#L310

Possible fix: Can a configurable port name be added?

The render:
```
---
# Source: retool/templates/deployment_workflows.yaml
apiVersion: v1
kind: Service
metadata:
  name: my-retool-workflow-backend
  namespace: my-retool
spec:
  selector:
    retoolService: my-retool-workflow-backend
  ports:
    - protocol: TCP
      port: 80
      targetPort: 3000
---
```

# Error 2 and 3

port names are unable to follow convention

https://github.com/tryretool/retool-helm/blob/main/charts/retool/templates/deployment_workflows_worker.yaml#L234-L239

Possible fix: Can these port names be configurable?

The render:
```
---
# Source: retool/templates/deployment_workflows_worker.yaml
apiVersion: v1
kind: Service
metadata:
  name: my-retool-workflow-worker
  namespace: my-retool
spec:
  selector:
    retoolService: my-retool-workflow-worker
  ports:
    - protocol: TCP
      port: 3005
      targetPort: 3005
      name: retool
    - protocol: TCP
      port: 9090
      targetPort: metrics
      name: metrics
---
```

My proposed fixes above should not be breaking changes.

Thanks

alexlo03:
Note that chart v5.0.10 works for us which we will continue to use for the time being.

PR #170, opened 2024-05-20
update port names for Services for istio compliance

avimoondra:
Addresses https://github.com/tryretool/retool-helm/issues/144, alternate to https://github.com/tryretool/retool-helm/pull/151

Addresses: https://github.com/tryretool/retool-helm/issues/155, alternate to https://github.com/tryretool/retool-helm/pull/156

Lighter weight approach than https://github.com/tryretool/retool-helm/pull/154

https://istio.io/v1.0/docs/setup/kubernetes/spec-requirements/

> Named ports: Service ports must be named. The port names must be of the form <protocol>[-<suffix>] with http, http2, grpc, mongo, or redis as the <protocol> in order to take advantage of Istio's routing features. For example, name: http2-foo or name: http are valid port names, but name: http2foo is not. If the port name does not begin with a recognized prefix or if the port is unnamed, traffic on the port will be treated as plain TCP traffic (unless the port [explicitly](https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service) uses Protocol: UDP to signify a UDP port).

https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/, could also use appProtocol if prefix in names is not desirable.

Tested in internal (balloon) instances.

(also fixes https://github.com/tryretool/retool-helm/issues/155)

### DIFF
--- a/charts/retool/charts/retool-temporal-services-helm/templates/server-deployment.yaml
+++ b/charts/retool/charts/retool-temporal-services-helm/templates/server-deployment.yaml
@@ -195,7 +195,7 @@ spec:
             - name: rpc
               containerPort: {{ include (printf "temporal.%s.grpcPort" $service) $ }}
               protocol: TCP
-            - name: metrics
+            - name: tcp-metrics
               containerPort: 9090
               protocol: TCP
           {{- if and (ne $service "worker") $.Values.server.useLegacyHealthProbe }}
--- a/charts/retool/charts/retool-temporal-services-helm/templates/server-deployment.yaml
+++ b/charts/retool/charts/retool-temporal-services-helm/templates/server-deployment.yaml
@@ -195,7 +195,7 @@ spec:
             - name: rpc
               containerPort: {{ include (printf "temporal.%s.grpcPort" $service) $ }}
               protocol: TCP
-            - name: tcp-metrics
+            - name: http-metrics
               containerPort: 9090
               protocol: TCP
           {{- if and (ne $service "worker") $.Values.server.useLegacyHealthProbe }}
--- a/charts/retool/charts/retool-temporal-services-helm/templates/server-service.yaml
+++ b/charts/retool/charts/retool-temporal-services-helm/templates/server-service.yaml
@@ -66,9 +66,9 @@ spec:
       protocol: TCP
       name: grpc-rpc
     - port: 9090
-      targetPort: tcp-metrics
+      targetPort: http-metrics
       protocol: TCP
-      name: tcp-metrics
+      name: http-metrics
   selector:
     app.kubernetes.io/name: {{ include "temporal.name" $ }}
     app.kubernetes.io/instance: {{ $.Release.Name }}
--- a/charts/retool/charts/retool-temporal-services-helm/templates/server-service.yaml
+++ b/charts/retool/charts/retool-temporal-services-helm/templates/server-service.yaml
@@ -66,9 +66,9 @@ spec:
       protocol: TCP
       name: grpc-rpc
     - port: 9090
-      targetPort: metrics
+      targetPort: tcp-metrics
       protocol: TCP
-      name: metrics
+      name: tcp-metrics
   selector:
     app.kubernetes.io/name: {{ include "temporal.name" $ }}
     app.kubernetes.io/instance: {{ $.Release.Name }}
--- a/charts/retool/templates/_helpers.tpl
+++ b/charts/retool/templates/_helpers.tpl
@@ -306,6 +306,14 @@ Set code executor service name
 {{- end -}}
 
 {{/*
+Set multiplayer service name
+*/}}
+{{- define "retool.multiplayer.name" -}}
+{{ template "retool.fullname" . }}-multiplayer-ws
+{{- end -}}
+
+
+{{/*
 Set code executor image tag
 Usage: (template "retool.codeExecutor.image.tag" .)
 */}}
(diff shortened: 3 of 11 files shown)


charts/retool/templates/deployment_telemetry.yaml
```diff
@@ -175,7 +175,7 @@ metadata:
     {{- include "retool.telemetry.labels" . | nindent 4 }}
 spec:
   ports:
-    - name: statsd-udp
```
avimoondra, line 178: Not needed, but just doing for consistency 

> Non-TCP based protocols, such as UDP, are not proxied. These protocols will continue to function as normal, without any interception by the Istio proxy but cannot be used in proxy-only components such as ingress or egress gateways.

https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/

charts/retool/templates/_helpers.tpl
```diff
@@ -305,6 +305,14 @@ Set code executor service name
 {{ template "retool.fullname" . }}-code-executor
 {{- end -}}
 
+{{/*
+Set multiplayer service name
+*/}}
+{{- define "retool.multiplayer.name" -}}
+{{ template "retool.fullname" . }}-multiplayer-ws
+{{- end -}}
```
avimoondra, line 313: was using this originally to name ports, but port names can be max 15 characters

https://stackoverflow.com/questions/73330773/is-there-any-way-to-disable-or-increase-port-name-length-in-kubernetes

charts/retool/templates/deployment_backend.yaml
```diff
@@ -241,7 +241,7 @@ spec:
         {{- end }}
         ports:
         - containerPort: {{ .Values.service.internalPort }}
-          name: {{ template "retool.name" . }}
+          name: http-server
           protocol: TCP
 {{- if .Values.livenessProbe.enabled }}
         livenessProbe:
```
avimoondra, line 247: there is no service defined here - attempted to add, but wasn't able to get it to compile template w/out error 👀

roberto-retool (approved):

charts/retool/charts/retool-temporal-services-helm/templates/server-deployment.yaml
```diff
@@ -198,7 +198,7 @@ spec:
             - name: membership
               containerPort: {{ include (printf "temporal.%s.membershipPort" $service) $ }}
               protocol: TCP
-            - name: metrics
+            - name: http-metrics
```
roberto-retool, line 201: don't think we need this since this is the pod's port

charts/retool/charts/retool-temporal-services-helm/templates/server-service.yaml
```diff
@@ -87,9 +87,9 @@ spec:
       protocol: TCP
       name: grpc-membership
     - port: 9090
-      targetPort: metrics
+      targetPort: http-metrics
       protocol: TCP
-      name: metrics
+      name: http-metrics
```
roberto-retool, line 92: would change this if also change above comment

charts/retool/templates/_helpers.tpl
ryanartecona: IMO it could just be `multiplayer-ws` as port name, since it will already be on a Service object which has the retool.fullname namespace in its own name?

charts/retool/charts/retool-temporal-services-helm/templates/server-service.yaml
ryanartecona: does changing it complicate anything? I was wondering the same, but OTOH I don't have the change if only for consistency of naming scheme, if it doesn't hurt anything.

ryanartecona (approved):

charts/retool/charts/retool-temporal-services-helm/templates/server-service.yaml
roberto-retool: maybe a very minor downtime if service port changes before new pod comes up?
but yeah agree if we make them consistent we should do for all (e.g. the two ports defined above)

ryanartecona: oh right, we're using a named targetPort here. good point.

ryanartecona: actually I think technically, if this causes minor downtime, it's downtime either way, since we _have_ to update `targetPort` to match the pod's new port name which comes from the Deployment template. we could choose to or not to update `name` here on the Service object to match purely for consistency, but I don't think that would have any downtime since the Service port number (`port` here) doesn't change. I think?

charts/retool/templates/_helpers.tpl
avimoondra: i think the service portName just had to have http- prefix, but will just keep as is for consistency

and then you can have "retool.name" be longer than 15 chars i suppose 😅

charts/retool/charts/retool-temporal-services-helm/templates/server-service.yaml
avimoondra: it seems like the ports that only need naming are on service and pod 

https://istio.io/v1.0/docs/setup/kubernetes/spec-requirements/

I think will stick with the current changes then 👍

avimoondra:
@alexlo03 Re-visiting this, I think it should work istio compliance requirements?

alexlo03 (approved):
LGTM, I think this is a better/cleaner/clearer approach 🙏 

I do not know what the upgrade looks like for existing deployments - renaming a port "should be a no-op" but IDK for sure.

PR #186, opened 2024-09-13
Scale Jobs-runner resources separately from main_backend 

kcheng16:
Added option to set resource.request/limit for jobs-runner and workflows-backend container, separating it from same resources used in for main-backend container

### DIFF
--- a/charts/retool/values.yaml
+++ b/charts/retool/values.yaml
@@ -348,7 +348,7 @@ workflows:
     # will launch 9 pods -- 4 workflow backend, 1 workflow workers, and 4 for temporal cluster
     replicaCount: 1
 
-      # If necessary, specify the resources to provision the workflows-backend pod separately from the main backend pods.
+    # If necessary, specify the resources to provision the workflows-backend pod separately from the main backend pods.
     # resources:
     #   limits:
     #     cpu: 4096m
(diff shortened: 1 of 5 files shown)


charts/retool/values.yaml
```diff
@@ -231,7 +231,7 @@ extraVolumeMounts: []
 
 extraVolumes: []
 
-# These resource specifications apply to the main backend and workflows backend pods.
+# These resource specifications will apply to the main backend, jobs-runner, db-connector, and workflows backend pods(unless if container specific resources are set)
```
JatinNanda, line 234: nit, reword:

# These resource specifications will apply to the main backend, jobs-runner, dbconnector, and workflows-backend pods unless container specific resources are set.

charts/retool/values.yaml
```diff
@@ -299,6 +299,15 @@ jobRunner:
   # will already launch a job runner pod
   # enabled: true
 
+  # These resource apply only to the jobs-runner pod. Set the resources to scale jobs-runner pod separately from main-backend pods.
```
JatinNanda, line 302: nit reword:

If necessary, specify the resources to provision the jobRunner pod separately from the main backend pods.

JatinNanda (approved):

PR #190, opened 2024-11-09
[fix] Update chart to handle the opt-in mechanism for Java dbconnecto…

jjlgao:
…r before 3.93-edge

In chart version 6.2.9, we changed Java dbconnector to be opt out, using a new system that uses an env var so that it's opt out even outside of Helm. This PR adds back in the original functionality for Java dbconnector opt-in logic (although now it will be opt-out due to the change in the values file default), so that customers on older versions of Retool won't see regressions as they upgrade their helm chart version.


KnlnKS (approved):

charts/retool/templates/deployment_backend.yaml
```diff
@@ -77,21 +77,29 @@ spec:
             value: {{ template "retool.deploymentTemplateVersion" . }}
           - name: NODE_ENV
             value: production
-          {{- if include "retool.jobRunner.enabled" . }}
-          {{ if ( not $.Values.dbconnector.java.enabled ) }}
-          - name: DISABLE_JAVA_DBCONNECTOR
-            value: "true"
-          {{ end }}
+          {{ if and ( or (eq "latest" $.Values.image.tag ) ( semverCompare ">= 3.93.0-0" $.Values.image.tag ) ) ( .Values.jobRunner.enabled ) }}
```
anna-yn, line 80: i think `semverCompare` might give you trouble with image tags that are like `3.xx.x-stable` or `-edge`

jjlgao: I tested against the `-edge` ones and it works fine. I think it supports them because it's supposed to support SHAs there. That's why the comparison is against `3.93.0-0`.

I might add extra tests against a helm chart that has those versions just in case!

charts/retool/templates/deployment_backend.yaml
```diff
@@ -77,21 +77,29 @@ spec:
...
+          - name: SERVICE_TYPE
+            value: MAIN_BACKEND,DB_CONNECTOR,DB_SSH_CONNECTOR
+          {{ else if and ( not or (eq "latest" $.Values.image.tag ) ( semverCompare ">= 3.93.0-0" $.Values.image.tag ) ) ( not .Values.jobRunner.enabled ) ( $.Values.dbconnector.java.enabled ) }}
+          - name: SERVICE_TYPE
+            value: MAIN_BACKEND,DB_CONNECTOR,DB_SSH_CONNECTOR,JOBS_RUNNER,JAVA_DBCONNECTOR
+          {{ else }}
+          - name: SERVICE_TYPE
+            value: MAIN_BACKEND,DB_CONNECTOR,DB_SSH_CONNECTOR,JOBS_RUNNER
+          {{ end }}
+          {{ if and ( or (eq "latest" $.Values.image.tag ) ( semverCompare ">= 3.93.0-0" $.Values.image.tag ) ) ( not $.Values.dbconnector.java.enabled ) }}
```
anna-yn, line 99: the long if list feels like trouble, esp if we want to change how we construct these in the future (which we'll certainly do, esp with splitting out main backend and dbc)..i wonder if we could do something like
`value: MAIN_BACKEND{{ if <should have dbc> ",DB_CONNECTOR" else "" }}{{if <should have java dbc> ",JAVA_DBCONNECTOR" else "" }}

jjlgao: uhhhh maybe that's a good idea let me try it

charts/retool/templates/deployment_workflows.yaml
```diff
@@ -66,12 +66,28 @@ spec:
...
+          */}}
+          {{- if and ( not $retool_version_with_java_dbconnector_opt_out ) ( $.Values.dbconnector.java.enabled ) }}
+            {{- $serviceType = append $serviceType "JAVA_DBCONNECTOR" }}
+          {{- end }}
+          {{- /*
+            It may seem counterintuitive to add the JOBS_RUNNER service type only without a jobs runner.
+            The reason for this is that the backend needs to act as a jobs runner, if the jobs runner is not enabled.
+          */}}
+          {{- if not .Values.jobRunner.enabled }}
+            {{- $serviceType = append $serviceType "JOBS_RUNNER" }}
```
anna-yn, line 83: wait, the workflow backend should never have the jobs runner service. it wasn't in the original list either

jjlgao: whoops

charts/retool/templates/deployment_backend.yaml
```diff
@@ -77,21 +77,28 @@ spec:
...
+          {{- $retool_version_with_java_dbconnector_opt_out := ( or (eq "latest" $.Values.image.tag ) ( semverCompare ">= 3.93.0-0" $.Values.image.tag ) ) }}
+          {{- $serviceType := list "MAIN_BACKEND" "DB_CONNECTOR" "DB_SSH_CONNECTOR" }}
+          {{- /*
+            JAVA_DBCONNECTOR in the service type only applies before the version of Retool that changes it to opt-out (3.93.0-edge),
+            and only if the Java dbconnector is enabled in values.yaml.
+          */}}
+          {{- if and ( not $retool_version_with_java_dbconnector_opt_out ) ( $.Values.dbconnector.java.enabled ) }}
+            {{- $serviceType = append $serviceType "JAVA_DBCONNECTOR" }}
+          {{- end }}
+          {{- /*
```
anna-yn, line 89: love the comments!

PR #191, opened 2024-11-19
[INF-4851] ESO external secret version

slaminad:
set `.Values.externalSecrets.externalSecretsOperator.useLegacyCR` to `false` to use ESO CRDs


charts/retool/templates/externalsecret.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.externalSecrets.externalSecretsOperator.enabled }}
+{{- if and .Values.externalSecrets.externalSecretsOperator.enabled (not .Values.externalSecrets.externalSecretsOperator.esoVersion) }}
```
JatinNanda, line 1: esoVersion is the flag to forgo legacy behavior? seems like a potentially confusing name since version doesn't sound like a boolean

slaminad: yea i was trying to think of a good variable name for this. do you have any suggestions?

i also wasn't thinking of exposing it in values.yaml just bc we're using it for testing it atm

slaminad: maybe we could go the full `.Values.externalSecrets.externalSecretsOperator.esoVersion.enabled` route

JatinNanda: i was thinking just force this new version and make it a major version upgrade instead of branching 🤔

slaminad: i'm fine with that but i'll have to test that migrating between the two looks good

JatinNanda: if that's too much work, i'd say use a variable that's set to true by default that's like 'useLegacyCR'

JatinNanda (approved):

PR #196, opened 2024-11-22
[fix] Retool internal versions are not semver compliant, so ensure that our helm chart is able to support all version formats

jjlgao:


charts/retool/templates/_helpers.tpl
```diff
@@ -200,6 +200,8 @@ Usage: (include "retool.workflows.enabled" .)
 */}}
 {{- define "retool.workflows.enabled" -}}
 {{- $output := "" -}}
+{{- $valid_retool_version_regexp := "([0-9]+]\\.[0-9]+(\\.[0-9]+)?(-[a-zA-Z0-9]+)?)" }}
```
june-xu, line 203: what if the retool version is `pr-3.122.0-9492dbc` for an internal instance?

charts/retool/templates/deployment_backend.yaml
```diff
@@ -77,21 +77,33 @@ spec:
...
-          {{ if ( not $.Values.dbconnector.java.enabled ) }}
-          - name: DISABLE_JAVA_DBCONNECTOR
-            value: "true"
-          {{ end }}
+          {{- /*
+            We are making an assumption here that if we see an invalid tag, we are running a relatively recent version of Retool.
+            Older versions of Retool should always have a semver-compatible tag; only internal versions should be noncompliant.
+          */}}
+          {{- $valid_retool_version_regexp := "([0-9]+]\\.[0-9]+(\\.[0-9]+)?(-[a-zA-Z0-9]+)?)" }}
+          {{- $retool_version_with_java_dbconnector_opt_out := ( or ( not ( regexMatch $valid_retool_version_regexp $.Values.image.tag ) ) ( semverCompare ">= 3.93.0-0" ( regexFind $valid_retool_version_regexp $.Values.image.tag ) ) ) }}
```
june-xu, line 85: reading this is tough in line, is there a helper file we can add this to lol

june-xu: an example of a helm function: https://github.com/tryretool/retool-k8s/blob/e053bbf66ad991dbed245adf745aa4605cded79d/terragrunt/src/managed-self-hosted/helm/retool-fork/charts/postgresql/charts/common/templates/_ingress.tpl#L36-L42 we can put it in a file called `_helpers.tpl`

mertbozfakioglu (commented):
This LGTM but I'd need to defer to someone from infra for the approval

.github/workflows/ci.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
           version: v3.6.3
       - uses: actions/setup-python@v4
         with:
-          python-version: 3.7
+          python-version: 3.12
```
jjlgao, line 30: this is because github changed the ubuntu OS so now only python 3.9-3.12 are supported

june-xu (approved):
nice! thanks for the refactor

PR #204, opened 2025-01-16
Optional standalone dbconnector

ryanartecona:
Allow dbconnector to run as a standalone deployment for scale & resource flexibility for large deployments.

### DIFF
--- a/charts/retool/ci/standalone-dbconnector-enabled-option.yaml
+++ b/charts/retool/ci/standalone-dbconnector-enabled-option.yaml
@@ -1,0 +1,2 @@
+dbconnector:
+  enabled: true
--- a/charts/retool/templates/deployment_dbconnector.yaml
+++ b/charts/retool/templates/deployment_dbconnector.yaml
@@ -43,7 +43,7 @@ spec:
         readinessProbe:
           httpGet:
             path: {{ .Values.readinessProbe.path }}
-            port: 3002
+            port: {{ .Values.dbconnector.port }}
           initialDelaySeconds: {{ .Values.readinessProbe.initialDelaySeconds }}
           timeoutSeconds: {{ .Values.readinessProbe.timeoutSeconds }}
           successThreshold: {{ .Values.readinessProbe.successThreshold }}
@@ -51,7 +51,7 @@ spec:
         livenessProbe:
           httpGet:
             path: {{ .Values.livenessProbe.path }}
-            port: 3002
+            port: {{ .Values.dbconnector.port }}
           initialDelaySeconds: {{ .Values.livenessProbe.initialDelaySeconds }}
           timeoutSeconds: {{ .Values.livenessProbe.timeoutSeconds }}
           failureThreshold:  {{ .Values.livenessProbe.failureThreshold }}
@@ -265,6 +265,6 @@ spec:
   ports:
   - protocol: TCP
     name: http-server
-    port: 3002
-    targetPort: 3002
+    port: {{ .Values.dbconnector.port }}
+    targetPort: {{ .Values.dbconnector.port }}
 {{- end }}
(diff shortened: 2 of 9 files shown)


charts/retool/ci/standalone-dbconnector-enabled-option.yaml
```diff
@@ -0,0 +1,2 @@
+dbconnector:
```
jjlgao, line 1: glad to see this feature of testing with random extra charts was useful lol

values.yaml
```diff
@@ -428,11 +428,47 @@ workflows:
...
+  # resources configured for main backend.
+  # resources:
+  #   limits:
+  #     cpu: 4096m
+  #     memory: 8192Mi
+  #   requests:
+  #     cpu: 2048m
+  #     memory: 4096Mi
+
+  config:
```
jjlgao, line 463: did you mean to set this config intentionally? I'm surprised to see such a high default postgres max pool size, especially on the dbc deployment, as that would be specific to the Postgres connector which isn't global.

charts/retool/templates/deployment_workflows.yaml
```diff
@@ -66,16 +66,29 @@ spec:
...
+          {{- if not $.Values.dbconnector.enabled -}}
+            {{- $serviceType = concat $serviceType (list "DB_CONNECTOR" "DB_SSH_CONNECTOR") -}}
+          {{- end }}
           - name: SERVICE_TYPE
             value: {{ join "," $serviceType }}
+          {{- if $.Values.dbconnector.enabled }}
+          - name: DB_CONNECTOR_HOST
+            value: http://{{ template "retool.fullname" . }}-dbconnector
+          - name: DB_CONNECTOR_PORT
+            value: '3002'
```
jjlgao, line 90: nit: would consider making this configurable with a default of 3002, unless we hardcode the backend port as well.

charts/retool/templates/deployment_dbconnector.yaml
```diff
@@ -0,0 +1,270 @@
...
+kind: Service
+metadata:
+  name: {{ include "retool.dbconnector.name" . }}
+spec:
+  selector:
+    retoolService: {{ include "retool.dbconnector.name" . }}
+  ports:
+  - protocol: TCP
+    name: http-server
+    port: 3002
```
jjlgao, line 268: nit: similarly to the env vars, would make this configurable if possible

charts/retool/templates/deployment_workflows.yaml
```diff
@@ -66,16 +66,29 @@ spec:
...
+            If it is not enabled (default), the main backend acts as its own dbconnector.
+          */ -}}
+          {{- if not $.Values.dbconnector.enabled -}}
+            {{- $serviceType = concat $serviceType (list "DB_CONNECTOR" "DB_SSH_CONNECTOR") -}}
+          {{- end }}
           - name: SERVICE_TYPE
             value: {{ join "," $serviceType }}
+          {{- if $.Values.dbconnector.enabled }}
+          - name: DB_CONNECTOR_HOST
+            value: http://{{ template "retool.fullname" . }}-dbconnector
```
jjlgao, line 88: this would point dbc at the generic dbc deployment -- is it intentional that we aren't creating a workflows-dbconnector deployment here?

ryanartecona: it's intentional, yes, but not unchangeable. I think splitting dbconnector out of main backend is likely the main scalability unlock for large single-tenant deployments, and we're only just now adding it to retool-helm. going even further and splitting out a workflows-dbconnector from the main dbconnector seems premature, given that workflows doesn't need its own dbconnector to function and we do that mainly for scale reasons in our multi-tenant cloud.

charts/retool/templates/deployment_workflows.yaml
ryanartecona: done! I couldn't do it exactly how backend does it, but it's at least configurable.

slaminad (approved):
lgtm. tested on athelas

slaminad:
since we saw that the msh athelas was working the other day, i think this should be good to go?

Issue #199, opened 2024-12-17
Multiplayer/Ingress force retool as backend


davidaparicio:
Hello!

I deployed the latest stable version (3.114.2-stable) with the penultimate version of the [HELM Chart](https://artifacthub.io/packages/helm/retool/retool) (retool-6.2.14).

But the chart can't permit to redirect the path `/api/multiplayer/` to the serviceName: `retool-multiplayer-ws`.

In fact, the lines [40/44](https://github.com/tryretool/retool-helm/blob/main/charts/retool/templates/ingress.yaml#L40) have the full name of the application so it's not possible to overwrite to another value (in order to target the multiplayer webservice). :-/

Kind regards,
David

davidaparicio:
An example for an opensource helm chart:
https://github.com/codecentric/helm-charts/blob/master/charts/keycloak/values.yaml#L308

davidaparicio:
I retried today without sucess..
I deployed the latest stable version (3.114.5-stable) with the ultimate version of the [HELM Chart](https://artifacthub.io/packages/helm/retool/retool) (retool-6.3.0).

davidaparicio:
Hi @jjlgao,

Yes, I can share you the current overview on our retool namespace
```
❯ k get pods,deployments,services,ingress,secrets,configmaps -n retool
NAME                                          READY   STATUS    RESTARTS   AGE
pod/retool-9yy-8xx                            1/1     Running   0          150m
pod/retool-9yy-rxx                            1/1     Running   0          152m
pod/retool-code-executor-7d56-9bn4t           1/1     Running   0          152m
pod/retool-jobs-runner-9b9c-tcrn6             1/1     Running   0          152m
pod/retool-multiplayer-ws-6468-v78m9          1/1     Running   0          139m
pod/retool-workflow-backend-b9bc8-b7d57       1/1     Running   0          152m
pod/retool-workflow-worker-698b-hpvll         1/1     Running   0          152m

NAME                                      READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/retool                    2/2     2            2           261d
deployment.apps/retool-code-executor      1/1     1            1           261d
deployment.apps/retool-jobs-runner        1/1     1            1           261d
deployment.apps/retool-multiplayer-ws     1/1     1            1           139m
deployment.apps/retool-workflow-backend   1/1     1            1           261d
deployment.apps/retool-workflow-worker    1/1     1            1           261d

[...]

PR #205, opened 2025-01-21
[fix] Add default multiplayer ingress rule

jjlgao:
Addresses https://github.com/tryretool/retool-helm/issues/199, aims to automatically create an ingress rule, if multiplayer is enabled, to point requests for `/api/multiplayer` to the multiplayer pod. To avoid conflicts with folks who already have this ingress rule set, this automatic rule can be disabled using `multiplayer.ingress.enabled=false`.

Tested by applying to our test cluster, which created a new (inaccessible) ingress, which looks like this (redacted hostnames just in case, but they shouldn't be sensitive):
Screenshot 2025-01-22 at 8 15 26 AM
Screenshot 2025-01-22 at 8 15 34 AM

I also added these values to the `values.yaml` file:
```
ingress:
  enabled: true
  labels: {}
  annotations: {}
  # kubernetes.io/ingress.class: nginx
  # kubernetes.io/tls-acme: "true"
  hosts:
  - host: retool-dev.[hostname]
    paths:
      - path: /
  tls:
  - secretName: retool-dev-tls
    hosts:
      - retool-dev.[hostname]
  pathType: ImplementationSpecific
multiplayer:
  enabled: true
```

### DIFF
--- a/charts/retool/values.yaml
+++ b/charts/retool/values.yaml
@@ -461,6 +461,13 @@ multiplayer:
   # Labels for multiplayer pods
   labels: {}
 
+  # Paths to route to multiplayer pods; defaults to /api/multiplayer. Can specify both path and port.
+  ingress:
+    enabled: true
+    paths:
+      - path: /api/multiplayer
+        port: 80
+
 codeExecutor:
   # enabled by default from Chart version 6.0.15 and Retool image 3.20.15 onwards
   # explicitly set other fields as needed
--- a/charts/retool/values.yaml
+++ b/charts/retool/values.yaml
@@ -463,6 +463,7 @@ multiplayer:
 
   # Paths to route to multiplayer pods; defaults to /api/multiplayer. Can specify both path and port.
   ingress:
+    # This conditional is dependent on multiplayer.enabled.
     enabled: true
     paths:
       - path: /api/multiplayer
--- a/charts/retool/Chart.yaml
+++ b/charts/retool/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 6.3.5
+version: 6.3.6
 maintainers:
   - name: Retool Engineering
     email: engineering+helm@retool.com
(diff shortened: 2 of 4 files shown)


charts/retool/values.yaml
```diff
@@ -459,6 +459,14 @@ multiplayer:
   # Labels for multiplayer pods
   labels: {}
 
+  # Paths to route to multiplayer pods; defaults to /api/multiplayer. Can specify both path and port.
+  ingress:
+    # This conditional is dependent on multiplayer.enabled.
+    enabled: true
```
lukefoster11, line 465: I know this can be set to `false` but by default could this be a breaking change for anyone who has already set up this ingress rule?

lukefoster11 (approved):
Looks good, just had one question before merge

charts/retool/values.yaml
jjlgao: It will be, which I considered and decided to keep this way. I think there's no real way to make this default-true moving forward, without breaking folks who have the ingress rule already. I believe that what will happen is that people will run into an error on helm upgrade due to the duplicate routes, so when I tag the release for this PR I'll make a note in the release to expect the breaking change.

lukefoster11: Gotcha, yeah don't know a smooth alternative so this makes sense

charts/retool/values.yaml
```diff
@@ -459,6 +459,14 @@ multiplayer:
...
+    paths:
+      - path: /api/multiplayer
+        port: 80
```
alex-stabile, line 468: The multiplayer service uses 3009 as the default port, should that be set here?

jjlgao: No, this points at the entry port for the service, which is 80: https://github.com/tryretool/retool-helm/blob/main/charts/retool/templates/deployment_multiplayer_ws.yaml#L13

The destination port for the multiplayer deployment is hardcoded to 3001 here (probably should have been left at 3009 rather than introducing potential port conflicts, but that isn't relevant for this change), with another hardcoded env var pointing the container to port 3001: https://github.com/tryretool/retool-helm/blob/main/charts/retool/templates/deployment_multiplayer_ws.yaml#L78

alex-stabile (approved):

Issue #188, opened 2024-10-08
NodeAffinity for telemetry deployment


splio-rbechon:
nodeAffinity is not defined in deployment_telemetry.yml file.
I can't use it because this pod is unfortunately always scheduled on an arm64 node

Can you add nodeAffinity for telemetry deployment ? :pray: 

Have a nice day

jjlgao:
Hi! You should be able to define any extra sections you need for the telemetry deployment by using the `extraPodSpec` block.

PR #207, opened 2025-01-23
[feat] Adds nodeAffinity, nodeSelector, and tolerations blocks to tel…

jjlgao:
…emetry deployment

For https://github.com/tryretool/retool-helm/issues/188

### DIFF
--- a/charts/retool/Chart.yaml
+++ b/charts/retool/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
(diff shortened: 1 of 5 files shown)


ryanartecona (commented):
I don't think this is necessary? it looks like this would match the pattern established for the other deployments in here, but also I think all of affinity, nodeSelector, and tolerations could just be put into the existing extraPodSpec so this doesn't need to be added?

jjlgao:
Will close and comment on the original issue. Didn't realize the extraPodSpec was that flexible.

PR #208, opened 2025-01-24
allow external secrets to potentially clobber other secrets if opted in

JatinNanda:
if external secrets are currently enabled, they don't let us specify the regular configuration secrets as a safeguard from clobbering.

this flag should let us both use an externalsecret for something like the postgres password and just use a regular secret for the other things like license key

### DIFF
--- a/charts/retool/templates/_helpers.tpl
+++ b/charts/retool/templates/_helpers.tpl
@@ -351,3 +351,10 @@ Usage: (template "retool.codeExecutor.image.tag" .)
 {{- end -}}
 {{- $output -}}
 {{- end -}}
+
+{{/*
+Checks whether or not ExternalSecret definitions are enabled and can potentially clobber secrets or explicitly allow additional direct secret refs.
+*/}}
+{{- define "shouldIncludeConfigSecretsEnvVars" -}}
+{{- or (not (or (.Values.externalSecrets.enabled) (.Values.externalSecrets.externalSecretsOperator.enabled))) .Values.includeConfigSecrets -}}
+{{- end -}}
--- a/charts/retool/templates/deployment_backend.yaml
+++ b/charts/retool/templates/deployment_backend.yaml
@@ -153,7 +153,7 @@ spec:
 
           {{- include "retool.telemetry.includeEnvVars" . | nindent 10 }}
 
-          {{- if and (not .Values.externalSecrets.enabled) (not .Values.externalSecrets.externalSecretsOperator.enabled) }}
+          {{- if (include "shouldIncludeConfigSecretsEnvVars" .) }}
           - name: LICENSE_KEY
             valueFrom:
               secretKeyRef:
--- a/charts/retool/values.yaml
+++ b/charts/retool/values.yaml
@@ -95,6 +95,9 @@ externalSecrets:
   # Support for legacy external secrets, note this is deprecated in favour of External Secrets Operator: https://github.com/godaddy/kubernetes-external-secrets
   # This mode only allows a single secret name to be provided.
   enabled: false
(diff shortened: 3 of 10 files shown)


charts/retool/values.yaml
```diff
@@ -95,6 +95,8 @@ externalSecrets:
   # Support for legacy external secrets, note this is deprecated in favour of External Secrets Operator: https://github.com/godaddy/kubernetes-external-secrets
   # This mode only allows a single secret name to be provided.
   enabled: false
+  # If External Secrets are in use, determines whether or not to additionally support reading in configuration secrets as kubernetes secrets directly.
```
jjlgao, line 98: i feel like this param is not completely clear

i like the explanation u made in the description of the PR, maybe add the entire string to the comment here?

charts/retool/templates/deployment_backend.yaml
```diff
@@ -153,7 +153,7 @@ spec:
 
           {{- include "retool.telemetry.includeEnvVars" . | nindent 10 }}
 
-          {{- if and (not .Values.externalSecrets.enabled) (not .Values.externalSecrets.externalSecretsOperator.enabled) }}
+          {{- if and (not .Values.externalSecrets.enabled) (not .Values.externalSecrets.externalSecretsOperator.enabled) (not .Values.externalSecrets.allowOtherSecrets) }}
```
ryanartecona, line 156: I think this logic is wrong. shouldn't it be
```
if or (not (or (externalSecrets.enabled) (esOperator.enabled))) allowOtherSecrets
```
which is roughly "if neither externalSecret source is enabled, or if they are but allowOtherSecrets is true"

ryanartecona: also would love to put this in a helper at this point 😄

JatinNanda: yeah it's wrong, i did this right before bed, will fix and put into a helper

charts/retool/templates/_helpers.tpl
```diff
@@ -351,3 +351,10 @@ Usage: (template "retool.codeExecutor.image.tag" .)
 {{- end -}}
 {{- $output -}}
 {{- end -}}
+
+{{/*
+Checks whether or not ExternalSecret definitions are enabled and can potentially clobber secrets or explicitly allow additional direct secret refs.
+*/}}
+{{- define "checkExternalSecretsClobbering" -}}
```
ryanartecona, line 358: I would name this something a little more directly tied to what it gates, like `shouldIncludeConfigSecretsEnvVars` or smth

charts/retool/values.yaml
```diff
@@ -95,6 +95,9 @@ externalSecrets:
   # Support for legacy external secrets, note this is deprecated in favour of External Secrets Operator: https://github.com/godaddy/kubernetes-external-secrets
   # This mode only allows a single secret name to be provided.
   enabled: false
+  # If external secrets are currently enabled, it is disallowed to specify regular configuration secrets as a safeguard from clobbering.
+  # This flag allows bypassing that check and specifying both an ExternalSecret and a regular secret for different secrets.
+  allowOtherSecrets: false
```
ryanartecona, line 100: thinking naming...maybe `includeConfigSecrets`?

also, do we want to doc a recommendation towards true/false? is the `false` default what we'd usually want or just for backcompat?

ryanartecona (approved):

charts/retool/values.yaml
JatinNanda: yup, good idea. and yeah false the default just for backcompat

charts/retool/templates/_helpers.tpl
JatinNanda: agree

PR #209, opened 2025-01-28
Fix shouldIncludeConfigSecretsEnvVars output format

JatinNanda:
need string output for this to work ._. (was previously always evaluating to 'true' since it was a non-empty string)

tested it this time with all combinations of inputs :')


lukefoster11 (approved):
lgtm (looks _great_ to me)

PR #212, opened 2025-02-13
ExternalSecrets: Made SecretStoreRef and refreshInterval configurable via values.yaml

dmatviichuk:
- Made refreshInterval configurable via values.yaml.
- Introduced configurable SecretStoreRef parameters.
- Updated externalSecret.yaml to reference values from values.yaml.
- Improved Helm chart flexibility for external secrets management.

### DIFF
--- a/charts/retool/Chart.yaml
+++ b/charts/retool/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 6.4.1
+version: 6.4.2
 maintainers:
   - name: Retool Engineering
     email: engineering+helm@retool.com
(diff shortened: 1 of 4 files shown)


jjlgao:
Hi Dmytro, apologies as I just saw this. Just want to check before I take a look, did you have a use case you're aiming to solve for this change?

jjlgao (approved):

jjlgao:
> Hi @jjlgao, we use Vault instead of AWS Secrets Manager as the secretStoreRef. Additionally, Retool secrets do not change frequently, so we want the ability to configure the refreshInterval.

Makes sense, thank you! I'll at least merge this now, but we won't release a new version for this, unless you feel that it's urgent for you, in which case if you're able to change the chart version in Chart.yaml in this PR, I can do a new release for you.

PR #213, opened 2025-02-24
[fix] Reference global values when grabbing multiplayer name in ingress

jjlgao:
Tested this internally (msh test cluster), since this template reference is in a loop, we need to point it at $ instead of ., which points at the wrong context here.


lukefoster11 (approved):
lg

PR #214, opened 2025-02-28
docs: multiplayer env variable correction

davidaparicio:
As discussed in the issue #199, I found an error to add the environment variable as it was written ;-)

In fact, the environmentVariables map wasn't correct set

### DIFF
--- a/charts/retool/values.yaml
+++ b/charts/retool/values.yaml
@@ -452,8 +452,7 @@ multiplayer:
 
   # Set environment variables for multiplayer pods, e.g. defining which origin to use
   # environmentVariables:
-  #   - name: MAIN_DOMAIN
-  #     value: retool.foo.com
+  #   MAIN_DOMAIN: "retool.foo.com"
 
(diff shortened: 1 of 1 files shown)


davidaparicio:
The error was `cannot patch "retool-multiplayer-ws" with kind Deployment: Deployment.apps "retool-multiplayer-ws" is invalid: spec.template.spec.containers[0].env[13].name: Invalid value: "0": a valid environment variable name must consist of alphabetic characters, digits, '_', '-', or '.', and must not start with a digit (e.g. 'my.env-name',  or 'MY_ENV.NAME',  or 'MyEnvName1', regex used for validation is '[-._a-zA-Z][-._a-zA-Z0-9]*')`

The 13th value was the mine in my values.yaml, defined as it was in the documentation :)

```
environmentVariables:
- name: MAIN_DOMAIN
  value: retool.foo.com
```

jjlgao (approved):

Issue #217, opened 2025-03-06
Unable to attach an IRSA in Amazon EKS without manually patching ServiceAccount


wbhildeb:
With our self-hosted retool instance in Amazon EKS, we are unable to successfully attach an irsa to the retool service account without the following configurations:
```
env:
  AWS_ROLE_ARN: ${IRSA_ARN}
  AWS_WEB_IDENTITY_TOKEN_FILE: /var/run/secrets/eks.amazonaws.com/serviceaccount/token

serviceAccount:
  annotations:
    eks.amazonaws.com/role-arn: ${IRSA_ARN}
```

As well as manually patching the service account to set `automountServiceAccountToken: true`

wbhildeb:
```
diff --git a/charts/retool/templates/serviceaccount.yaml b/charts/retool/templates/serviceaccount.yaml
index ea8e3ac..90ec6ca 100644
--- a/charts/retool/templates/serviceaccount.yaml
+++ b/charts/retool/templates/serviceaccount.yaml
@@ -1,6 +1,7 @@
 {{- if .Values.serviceAccount.create }}
 apiVersion: v1
 kind: ServiceAccount
+automountServiceAccountToken: {{ .Values.serviceAccount.automountToken }}
 metadata:
   name: {{ include "retool.serviceAccountName" . }}
   labels: {{- include "retool.labels" . | nindent 4 }}
diff --git a/charts/retool/values.yaml b/charts/retool/values.yaml
index dfd9392..e29b0c5 100644
--- a/charts/retool/values.yaml
+++ b/charts/retool/values.yaml
@@ -205,6 +205,7 @@ serviceAccount:
   # If set and create is false, the service account must be existing
   name:
   annotations: {}
+  automountToken: false
 
 livenessProbe:
   enabled: true
diff --git a/values.yaml b/values.yaml
index dfd9392..e29b0c5 100644
--- a/values.yaml
+++ b/values.yaml
@@ -205,6 +205,7 @@ serviceAccount:
   # If set and create is false, the service account must be existing
   name:
   annotations: {}
+  automountToken: false
 
 livenessProbe:
   enabled: true
```

jjlgao:
Hi! It looks like this was solved in your PR #218 . Going to close this as it should be resolved, but feel free to reopen if not.

PR #218, opened 2025-03-07
feat(serviceaccount): add automountToken capabilities

wbhildeb:
As outlined in #217, we require this configuration to be able to attach an IRSA to the retool agent in our self-hosted retool instance on Amazon EKS.


ryanartecona:
thanks! this looks reasonable to me and it looks like tests passed. I'll release this shortly.

PR #221, opened 2025-03-11
Fixes #220

### LINKED ISSUE #220: Nil pointer error when deploying External Secrets with Helm and Terraform
When deploying the `retool` Helm chart using Terraform, the following error occurs during the plan phase:

```
Error: error running dry run for a diff: template: retool/templates/externalsecret.yaml:27:29: executing "retool/templates/externalsecret.yaml" at <.Values.externalSecrets.externalSecretsOperator.refreshInterval>: nil pointer evaluating interface {}.externalSecrets
```

We previously installed `retool` and encountered this issue when attempting to upgrade from version `6.3.6` to `6.4.2`.

We wanted to upgrade in order to set our own `secretStoreRef.name`, as it was previously hardcoded.

The issue occurs regardless of whether the `refreshInterval` value is explicitly set or not.

Our `values.yaml` configuration looks like this:

```yaml
externalSecrets:
  # Support for legacy external secrets, note this is deprecated in favour of External Secrets Operator: https://github.com/godaddy/kubernetes-external-secrets
  # This mode only allows a single secret name to be provided.
  enabled: false
  includeConfigSecrets: false
  name: retool-config
  externalSecretsOperator:
    enabled: true
    backendType: secretsManager
    secretStoreRef:
      name: retool
      kind: SecretStore
    secretRef:
      - name: retool-config
        path: ${sm_secret_name}
    # When true, uses kubernetes-client CRDs and not external-secrets CRDs
    useLegacyCR: false

extraManifests:
  - apiVersion: external-secrets.io/v1beta1
    kind: SecretStore
    metadata:
      name: retool
      annotations:
        helm.sh/hook-weight: "-4"
        helm.sh/hook: pre-install,pre-upgrade
    spec:
      provider:
        aws:
          service: SecretsManager
          region: ${sm_region}
          auth:
            jwt:
              serviceAccountRef:
                name: retool
```

We tried the following workarounds, but unfortunately, it didn't help:
- Explicitly setting `externalSecrets.externalSecretsOperator.refreshInterval` in `values.yaml`
- Using `set` in Terraform to define `externalSecrets.externalSecretsOperator.refreshInterval`

ryanartecona:


jjlgao (approved):
Thanks for watching the issues and doing this! Sorry, the `.Values` and `$.Values` thing has bitten me more than once now, I have to think about possibly enforcing usage of `$` if it's even possible to do.

PR #222, opened 2025-03-12
Allow configuring labels, annotations,  and ports for the multiplayer Service

abooij:
The main use for this is that in our implementation, the multiplayer Service is exposed as a Network Endpoint Group in Google Cloud. This requires adding a particular `annotation` to the Service. So I am using the following Service definition, managed outside of the Helm chart:
```yaml
apiVersion: v1
kind: Service
metadata:
  name: my-retool-multiplayer-ws
  annotations:
    cloud.google.com/neg: '{"exposed_ports": {"3001": {"name": "retool-multiplayer-neg"}}}'
spec:
  type: ClusterIP
  ports:
    - port: 3001
      targetPort: 3001
      protocol: TCP
      name: http
  selector:
    retoolService: retool-multiplayer-ws
```
(Our setup is roughly along the lines of [this guide](https://cloud.google.com/kubernetes-engine/docs/how-to/standalone-neg).)

It'd be amazing if this pattern could be supported through the Helm chart directly. It is already supported for the backend Service, simply by setting this in `values.yaml`:
```yaml
service:
  type: ClusterIP
  externalPort: 3000
  internalPort: 3000
  annotations:
    cloud.google.com/neg: '{"exposed_ports": {"3000": {"name": "retool-backend-neg"}}}'
  labels: {}
  selector: {}
```

I am not a Kubernetes expert, and haven't messed with Helm charts before. So it's quite likely something needs adjusting. Also, this code is untested as of writing. Feel free to push changes to this branch.

### DIFF
--- a/charts/retool/templates/deployment_multiplayer_ws.yaml
+++ b/charts/retool/templates/deployment_multiplayer_ws.yaml
@@ -3,14 +3,37 @@ apiVersion: v1
 kind: Service
 metadata:
   name: {{ template "retool.multiplayer.name" . }}
+  labels:
+    {{- include "retool.labels" . | nindent 4 }}
+  {{- if .Values.service.labels }}
+  {{- range $key, $value := .Values.service.labels }}
+    {{ $key }}: {{ $value | quote }}
+  {{- end }}
+  {{- end }}
+  {{- if .Values.multiplayer.service.annotations }}
+  {{- with .Values.multiplayer.service.annotations }}
+  annotations:
+    {{- range $key, $value := . }}
+      {{ $key }}: {{ $value | quote }}
+    {{- end }}
+  {{- end }}
+  {{- end }}
 spec:
   selector:
     retoolService: {{ template "retool.multiplayer.name" . }}
   ports:
   - name: http-server
     protocol: TCP
+    {{- if .Values.multiplayer.service.externalPort }}
+    port: {{ .Values.service.externalPort }}
+    {{- else }}
     port: 80
+    {{- end }}
+    {{- if .Values.multiplayer.service.internalPort }}
+    targetPort: {{ .Values.service.internalPort }}
+    {{- else }}
     targetPort: 3001
+    {{- end }}
 ---
 apiVersion: apps/v1
 kind: Deployment
--- a/values.yaml
+++ b/values.yaml
@@ -520,6 +520,12 @@ multiplayer:
       - path: /api/multiplayer
         port: 80
 
+  service:
+    externalPort: 80
+    internalPort: 3001
+    annotations: {}
+    labels: {}
+
 codeExecutor:
   # enabled by default from Chart version 6.0.15 and Retool image 3.20.15 onwards
   # explicitly set other fields as needed


charts/retool/templates/deployment_multiplayer_ws.yaml
```diff
@@ -3,14 +3,37 @@ apiVersion: v1
...
+    {{- end }}
+  {{- end }}
+  {{- end }}
 spec:
   selector:
     retoolService: {{ template "retool.multiplayer.name" . }}
   ports:
   - name: http-server
     protocol: TCP
+    {{- if .Values.multiplayer.service.externalPort }}
```
jjlgao, line 27: There's no need for an if/else here since you defined a default value for externalPort in values.yaml! Feel free to just have the `port: {{ .Values.service.externalPort }}` here.

values.yaml
```diff
@@ -520,6 +520,12 @@ multiplayer:
       - path: /api/multiplayer
         port: 80
 
+  service:
```
jjlgao, line 523: We have two values.yaml files unfortunately, and they need to match. Would you be able to copy this to the other one?

charts/retool/templates/deployment_multiplayer_ws.yaml
```diff
@@ -3,14 +3,37 @@ apiVersion: v1
...
+    port: {{ .Values.service.externalPort }}
+    {{- else }}
     port: 80
+    {{- end }}
+    {{- if .Values.multiplayer.service.internalPort }}
```
jjlgao, line 32: same here!

jjlgao:
Formatting generally looks good here; I may be able to help test this to make sure it works as expected. Would you be able to make the changes as suggested?

charts/retool/templates/deployment_multiplayer_ws.yaml
abooij: With this if/else, it's backwards compatible in case people don't have the `externalPort` in their current `values.yaml`. If we don't have it, then if people quietly upgrade their chart without adding to `values.yaml`, things would likely break.

abooij:
> Formatting generally looks good here; I may be able to help test this to make sure it works as expected. Would you be able to make the changes as suggested?
> 
> Unfortunately, I think it may be difficult for me to push directly to the branch as it's a fork (and I actually don't know how easy it would be for me to push to a fork).

I'm not sure what you mean with the difficulty. You can literally just checkout the branch and push to it.

charts/retool/templates/deployment_multiplayer_ws.yaml
jjlgao: Ah, thanks for the explanation! 

Apologies for being unclear, since you set a default value in values.yaml, it should carry over if they did not define it. I thought our default values.yaml would still act as the final fallback set of values in the event that someone didn't have a value set, but I'm hazy now and don't remember for sure.

Since you're leaving this check, I did notice that you've used `.Values.multiplayer.service.externalPort` as the if condition, but `.Values.service.externalPort` as the actual value being set. Should they match?

abooij: Oh yeah that was a mistake, it should've all referred to the `multiplayer` part of the config. Thanks for fixing it directly.

PR #231, opened 2025-04-09
[fix] Java dbconnector support when standalone dbconnector is enabled

jjlgao:
Currently, it's not possible to expose the Java dbconnector with the standalone dbconnector; this was an oversight when we merged that change.


ryanartecona (approved):
oh oops, thanks! I didn't realize these had to be specified separately.

PR #239, opened 2025-06-05
Expose more ExternalSecret options

ryanartecona:
Make all of the following possible 
* change the creationPolicy for an ExternalSecret
* change the deletionPolicy for an ExternalSecret
* mark the `secretRef` that uses an ExternalSecret itself optional, s.t. the pod is still allowed to start if that secret fails to sync or gets deleted


JatinNanda (approved):
👍

PR #241, opened 2025-06-13
add agents

lukeviens:
Adding 2 new (optional) workers for agents - agent-worker and agent-eval-worker
Refactoring workflows worker logic to be reusable across all Temporal workers
Refactoring workflows logic to move temporal to the top-level (independent from workflows)

### DIFF
--- a/charts/retool/Chart.yaml
+++ b/charts/retool/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 6.5.0
+version: 6.6.0
 maintainers:
   - name: Retool Engineering
     email: engineering+helm@retool.com


charts/retool/values.yaml
```diff
@@ -402,28 +402,6 @@ workflows:
...
-    # whether to use TLS/SSL when connecting to Temporal
-    # sslEnabled: false
-    # base64 encoded string of TLS/SSL client certificate to use for mTLS
-    # sslCert:
-    # base64 encoded string of TLS/SSL client certificate secret key to use for mTLS
-    # sslKey:
-    # recommended alternative to sslKey. Name and key of k8s secret containing base64 encoded sslKey
-    # sslKeySecretName:
-    # sslKeySecretKey
-
```
ryanartecona, line 426: moving this to the top level is a breaking change. any way we can avoid? can we just leave it here?

charts/retool/templates/_workers.tpl
```diff
@@ -0,0 +1,392 @@
+{{- define "retool.worker.deployment" -}}
+
+{{- $ := .context -}}
```
ryanartecona, line 3: I...didn't think you needed to thread context through like this? IIUC the `$` in helm is pre-set to the top level of the template rendering context object and available as a global in all partials & templates. so, plumbing in `"context" .` at the call site and then doing this here aren't actually doing anything? or rather they are doing what helm already does automatically? I think?

ryanartecona (commented):
to make sure these templates are linted & checked against k8s resource schemas, you should add a file in charts/retool/ci like `test-agents-enabled-option.yaml` that just has `agents: enabled: true`

charts/retool/templates/_workers.tpl
lukeviens: i think it actually requires that, since the template can only see what it receives (in this case a dict), so you need to hydrate $ again lol:

https://stackoverflow.com/questions/72956149/does-not-resolve-to-the-root-context-from-within-a-helm-template#:~:text=I%20pass%20a%20string%20value,However%20I%20get%20an%20error

charts/retool/values.yaml
lukeviens: yep, that would be going back to `templatize temporal workers`! the way in which I refactored it out shouldn't be breaking though, unless i've taken a wrong path? if I didn't handle a scenario (`temporal: enabled/disabled`, both in and/or out of `workflows:`, as well as handling `retool-temporal-services-helm:`), then that's definitely a clear mistake. i think it can certainly be cleaned up a bit tho (move logic -> `_helpers` better, there's some littered if statements)

lukeviens: or more concisely, the path of precedence is currently:
1. `workflows: temporal:`
2. `temporal:`
3. `retool-temporal-services-helm:`

charts/retool/templates/_workers.tpl
JatinNanda: i know i suggested this single template with includes approach for the deployment spec, but if they are really this different that you're having to pass in large dicts of stuff, imo you might as well just copy/paste the two specs

i think we only really get a readability/re-use benefit from templates/includes if we are able to like pretty cleanly generate the rendered specs with just like a prefix flag or something, but seems like there's enough differences here? i think the end result will be a lot easier to parse if they were just separate with some duplication

charts/retool/values.yaml
JatinNanda: i'd +1 to encapsulating this fallback logic into a helper function, it's a bit hard to read in-place

also, this kinda means until we make a major bump + breaking change, customers could specify temporal stuff in like 3 different places, but i guess that isn't _that_ bad, so this feels ok to me

JatinNanda: i'd +1 to encapsulating this fallback logic into a helper function, it's a bit hard to read in-place
also, this kinda means until we make a major bump + breaking change, customers could specify temporal stuff in like 3 different places, but i guess that isn't that bad, so this feels ok to me

charts/retool/templates/_workers.tpl
lukeviens: just pushed up a diff `re-refactor workers`, I think it's much nicer. might be able to squeeze some more prettiness out, but this feels quite close. let me know if there's anything else structural/etc. to change, have done _quite_ a bit of validation at this point so should hopefully be in a pretty good spot

charts/retool/values.yaml
lukeviens: done, should be encapsulated a bit better in _helpers.tpl now

charts/retool/templates/_workers.tpl
```diff
@@ -0,0 +1,413 @@
+{{- define "retool.workers" -}}
+- parent: agents
```
jjlgao, line 2: should this be defined in values.yaml, in case someone wants to spin up a subset of workers? Or do we expect folks to pretty much just need all of them every time

nit: I also see this variable is only used once below, should it maybe just go there instead of being duplicated

charts/retool/templates/_workers.tpl
```diff
@@ -0,0 +1,413 @@
...
+{{- $parent := .parent -}}
+{{- $workerType := .workerType -}}
+{{- $parentValues := index $.Values $parent -}}
+
+{{- $workerValues := $parentValues.worker -}}
+{{- if eq $workerType "agentEval" -}}
+  {{- $workerValues = $parentValues.evalWorker -}}
+{{- end -}}
+
+{{- $healthcheckPort := ternary 3012 3005 (eq $workerType "agentEval") -}}
```
jjlgao, line 32: why does the agentEvalWorker have a different healthcheck port than the rest? I believe we also normally allow folks to adjust these, as sometimes people don't have the right ports free. Is it possible to have defaults in values.yaml instead?

charts/retool/templates/_workers.tpl
```diff
@@ -0,0 +1,413 @@
...
+        {{- end }}
+        ports:
+        - containerPort: {{ $healthcheckPort }}
+          name: http-server
+          protocol: TCP
+        - containerPort: 9090
+          name: http-metrics
+          protocol: TCP
+
+{{- if $.Values.livenessProbe.enabled }}
```
jjlgao, line 306: do all of these agents have readiness/liveness probes matching backend's? if not, you're going to get crashes when they try to come up.

charts/retool/templates/_workers.tpl
```diff
@@ -0,0 +1,413 @@
...
+          mountPath: {{ .mountPath }}
+          subPath: {{ .subPath }}
+{{- end }}
+    {{- if $.Values.image.pullSecrets }}
+      imagePullSecrets:
+{{ toYaml $.Values.image.pullSecrets | indent 8 }}
+    {{- end }}
+    {{- if $.Values.affinity }}
+      affinity:
+{{ toYaml $.Values.affinity | indent 8 }}
```
jjlgao, line 354: hmm, also wondering about this I guess, what are the chances folks would want agent-specific affinity rules vs global ones. Probably not important yet?

jjlgao:
Looks pretty good, just a few nits overall. Will personally refrain from approving to allow the SHI team to do so!

charts/retool/templates/_workers.tpl
lukeviens: that's a really good question. the 2 scenarios that'd help right now are 1. `enabling agents without a workflow worker` and 2. `enabling e.g. agentWorker without agentEvalWorker`. I think 2 is probably not necessary(?), but 1 could be useful, since currently enabling agents forces you to have a workflow worker (as opposed to just the backend). tbh I don't know, what do you think?

charts/retool/templates/_workers.tpl
lukeviens: I definitely think likely enough to add in? just pushed a commit to add the overrides in either case !

PR #243, opened 2025-07-14
increase jobs runner default priority

lukeviens:
Adding two new default (higher) priority classes to the backend and jobs-runners pods, in order to eliminate crash loops during upgrades due to un-ran jobs (namely db migrations).

Currently, this change will only be applied when the global “priorityClassName” is not set. If it is, these default values will be overridden. In the future, it might be useful to create per-pod PriorityClass overrides in order to enable similar behavior, while still allowing a global priorityClassName.

### DIFF
--- a/charts/retool/templates/priorityclass.yaml
+++ b/charts/retool/templates/priorityclass.yaml
@@ -1,0 +1,23 @@
+{{- if not .Values.priorityClassName }}
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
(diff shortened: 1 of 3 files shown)


charts/retool/templates/priorityclass.yaml
```diff
@@ -0,0 +1,23 @@
+{{- if not .Values.priorityClassName }}
```
JatinNanda, line 1: worth a comment explaining why we're intentionally deprioritizing backend here

JatinNanda (approved):
nice

PR #244, opened 2025-07-21
lock code-executor version to match backend

slaminad:


charts/retool/values.yaml
```diff
@@ -363,10 +363,7 @@ ui:
   labels: {}
 
 workflows:
-  # enabled by default from Chart version 6.0.2 and Retool image 3.6.11 onwards
-  # explicitly set other fields as needed
-
-  # enabled: true
+  enabled: false
```
mikewang333, line 366: spoke offline but let's keep this `# enabled: true` since we do default to true

slaminad: done!

mikewang333 (approved):

PR #245, opened 2025-07-24
remove agents-preview note

lukeviens:
removing the agents-preview note, since it is now possible to enable agents via the edge release.


JatinNanda (approved):
huUge

PR #247, opened 2025-08-22
Add .Values.config.postgresPoolMaxSize

taylormusolf:
Remove hardcoded value for DBCONNECTOR_POSTGRES_POOL_MAX_SIZE with value: "100" in both _workers.tpl and deployment_workflows.yaml.  With this update users will need to set DBCONNECTOR_POSTGRES_POOL_MAX_SIZE in the values.yaml for it to get applied to workflows.

### DIFF
--- a/charts/retool/Chart.yaml
+++ b/charts/retool/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 6.7.0
+version: 6.8.0
 maintainers:
   - name: Retool Engineering
     email: engineering+helm@retool.com
--- a/charts/retool/templates/_workers.tpl
+++ b/charts/retool/templates/_workers.tpl
@@ -115,17 +115,8 @@ spec:
           - name: WORKER_TEMPORAL_TASKQUEUE
             value: {{ $taskqueue }}
           {{- end }}
-          {{- $envVars := default (list) .Values.environmentVariables }}
-          {{- $poolMaxSet := false }}
-          {{- range $envVars }}
-          {{- if eq .name "DBCONNECTOR_POSTGRES_POOL_MAX_SIZE" }}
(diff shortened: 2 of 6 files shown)


golfdish:
rather than requiring users to supply a value, we probably want to leave 100 as a default while allowing users to override it, like we do for `DBCONNECTOR_QUERY_TIMEOUT_MS` here: https://github.com/tryretool/retool-helm/pull/247/files#diff-ce58e50773488bf13409091c468325d84396844ea7f7fff711b3c0933200e983R102-R109

taylormusolf:
That is a good suggestion.  I will add a condition where if it is set in the values.yml, then it won't get set here.

golfdish:
@taylormusolf I actually found a better example of how it could be done elsewhere: https://github.com/tryretool/retool-helm/blob/415fb24399547704af723467b1f7cc4b91ff0dee/charts/retool/templates/deployment_dbconnector.yaml#L85-L86

this takes a [config value](https://github.com/tryretool/retool-helm/blob/415fb24399547704af723467b1f7cc4b91ff0dee/charts/retool/values.yaml#L471) from the `dbconnector` section. you could similarly make a variable for the pool size under the [workflows config](https://github.com/tryretool/retool-helm/blob/415fb24399547704af723467b1f7cc4b91ff0dee/charts/retool/values.yaml#L411) (currently empty) and reference it in the workflows templates, which would have the same effect of setting a changeable default.

golfdish:
ah, I missed the wrinkle of possibly double-applying the config and env values. I feel like maybe that could be avoided on the user's end? but I'll trust the testing you've done. ooc, does the current change achieve your goal of raising the dbconnector pool max size for workflows?

taylormusolf:
@golfdish The issue was that they needed to apply it to the backend container, and they were trying to do that via setting it in the environmentVariables, but that led to the issue with it being double-applied in workflows since those variables get blanket added to all containers.

PR #275, opened 2026-04-22
[INF-6675] add js executor

slaminad:


charts/retool/templates/deployment_js_executor.yaml
```diff
@@ -0,0 +1,189 @@
...
+{{- end }}
+      containers:
+      - name: js-executor
+        image: "{{ .Values.jsExecutor.image.repository }}:{{ include "retool.jsExecutor.image.tag" . }}"
+        imagePullPolicy: {{ .Values.image.pullPolicy }}
+        securityContext:
+          {{ if .Values.jsExecutor.securityContext }}
+{{ toYaml .Values.jsExecutor.securityContext | indent 10 }}
+          {{ else }}
+          privileged: true
```
mertbozfakioglu: ah we don't want this. it won't work with privileged. you need to give the seccomp profile and net_admin here. something like
```
securityContext:
          capabilities:
            add: ["NET_ADMIN"]
          seccompProfile:
            type: Localhost
            localhostProfile: {{ $.Values.jsExecutorDefault.seccompLocalhostProfile }}
```

charts/retool/values.yaml
```diff
@@ -606,6 +606,51 @@ codeExecutor:
...
+  volumeMounts: {}
+
+  # Config affinity and anti-affinity rules for the code executor pods
+  affinity: {}
+
+  # Resources for the JS executor. Most common issues will be seen with CPU usage as this will
+  # most likely be CPU bound. Adjust the CPU if latency increases under load.
+  resources:
+    limits:
+      cpu: 2000m
```
mertbozfakioglu, line 634: this is very very low. What is our code executor values rn?

charts/retool/values.yaml
```diff
@@ -606,6 +606,51 @@ codeExecutor:
...
+  # config: {
+  #   nodeOptions: --max_old_space_size=1024
+  # }
+  volumes: {}
+  volumeMounts: {}
+
+  # Config affinity and anti-affinity rules for the code executor pods
+  affinity: {}
+
+  # Resources for the JS executor. Most common issues will be seen with CPU usage as this will
```
mertbozfakioglu: It actually seems like it is memory bound more than CPU since we are transferring a lot of data through js executor

charts/retool/values.yaml
```diff
@@ -606,6 +606,51 @@ codeExecutor:
...
+  resources:
+    limits:
+      cpu: 2000m
+      memory: 2048Mi
+    requests:
+      cpu: 1000m
+      memory: 1024Mi
+
+  #dcm - should this be configurable?
+  # code executor uses nsjail to sandbox code execution. nsjail requires privileged container access.
```
mertbozfakioglu: same thing as before. privileged is a no-go

charts/retool/values.yaml
```diff
@@ -606,6 +606,51 @@ codeExecutor:
   securityContext:
     privileged: true
 
+# JS Executor
+jsExecutor:
+  # dcm - is it optional?
```
mertbozfakioglu: JS executor is a hard requirement. I don't think we need this configurable at all

charts/retool/values.yaml
```diff
@@ -606,6 +606,51 @@ codeExecutor:
...
+  enabled: true
+  # dcm - does this need to always align with the backend/CE image
```
mertbozfakioglu: could you clarify?

charts/retool/values.yaml
```diff
@@ -606,6 +606,51 @@ codeExecutor:
...
+  image:
+    repository: tryretool/js-executor-service
```
mertbozfakioglu, line 612: FYI I don't think we currently push this image to the public repo

charts/retool/values.yaml
```diff
@@ -606,6 +606,51 @@ codeExecutor:
...
+
+  replicaCount: 1
+
+  # Annotations for code executor pods
+  annotations: {}
+
+  # Labels for code executor pods
+  labels: {}
+
+  # Config for code executor. Node heap size limits can be overridden here
```
mertbozfakioglu: This is not true for JS Executor

values.yaml
```diff
@@ -606,6 +606,51 @@ codeExecutor:
   securityContext:
```
mertbozfakioglu, line 606: curious whats the difference between this file and charts/retool/values.yaml?

charts/retool/values.yaml
slaminad: actually all of this i just copied from the existing code executor values! what are your recommendations?

charts/retool/values.yaml
slaminad: yep just staging these for once we do

values.yaml
slaminad: there's no actual difference, i think our scripts require it to be in both places

charts/retool/values.yaml
slaminad: we have a [note](https://github.com/tryretool/retool-helm/blob/main/charts/retool/values.yaml#L566-L568) that for CE, we force the image to always align with the backend, mainly for security reasons

e.g. backend is `3.300.0`, CE cannot be anything other than `3.300.0`

i'm assuming we should enforce this same pattern for JS exec?

slaminad: the main consideration here is that if we ever build a one-off version for any of backend, CE, or JSE, we must also build a corresponding one-off version for the others

charts/retool/values.yaml
mertbozfakioglu: wow it is shocking that ce can run with 2gb of memory. I'd double it

charts/retool/values.yaml
mertbozfakioglu: I think it is good to enforce them to be in sync

charts/retool/templates/deployment_js_executor.yaml
```diff
@@ -0,0 +1,216 @@
...
+            type: Localhost
+            localhostProfile: {{ .Values.jsExecutor.seccompLocalhostProfile }}
+        env:
+          - name: DEPLOYMENT_TEMPLATE_TYPE
+            value: {{ template "retool.deploymentTemplateType" . }}
+          - name: DEPLOYMENT_TEMPLATE_VERSION
+            value: {{ template "retool.deploymentTemplateVersion" . }}
+          - name: NODE_ENV
+            value: production
+          - name: NODE_OPTIONS
```
mertbozfakioglu: this is not necessary

mertbozfakioglu (approved):
LGTM

PR #278, opened 2026-04-23
allow dynamic value references for env vars

ryanartecona:
there are some niche situations in which a user might want to split a helm release's values across separate values files and supply them separately. when helm merges separate values layers, objects merge cleanly (key name conflicts overwrite, but key names that don't conflict get unioned) but lists don't (a list in layer2 will clobber a list at the same path in layer1).

this makes it so the `env:` block in our values file, which is an object keyed by env var name, allows not only static string values but also dynamic reference values (such as a secretKeyRef), so that env values can be supplied across disparate layers without clobbering.

### DIFF
--- a/charts/retool/Chart.yaml
+++ b/charts/retool/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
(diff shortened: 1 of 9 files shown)


lukefoster11 (approved):
👍

PR #282, opened 2026-05-05
fix for bug in #278

ryanartecona:
should've more thoroughly tested :(


lukefoster11 (commented):
lint-chart gh action says chart version needs bump

lukefoster11 (approved):
👍

PR #307, opened 2026-06-05
rrGitServer: accept blob-storage env vars from .Values.env

JatinNanda:
Part of the **r2-cleanup** integration branch (merges into `r2-cleanup`, then `r2` via #303).

## Problem
`retool.rrGitServer.validateBlobStorage` only scanned `.Values.environmentVariables` and `.Values.environmentSecrets` for `RR_BLOB_STORAGE_PROVIDER` / `RR_DEFAULT_*`. Deployments that set those via the `.Values.env` map were forced to *duplicate* the `RR_*` vars into `environmentVariables` just to pass the check. And env vars supplied via `envFrom` (Secret/ConfigMap splat) are invisible at template time, so a valid setup using them had no way to satisfy the guard.

## Fix
1. Range over `.Values.env` (keyed by var name) in the same opt-out check.
2. Add `rrGitServer.skipBlobStorageValidation` (default `false`) as an explicit escape hatch for sources the chart cannot inspect at template time (e.g. `envFrom`). The failure message now points at this flag.

## Verification (`helm template`, rrGitServer.enabled=true)
| Case | Result |
|------|--------|
| `env.RR_BLOB_STORAGE_PROVIDER=gcs` | ✅ passes (previously required duplication) |
| `env.RR_DEFAULT_GCS_BUCKET=...` | ✅ passes |
| `skipBlobStorageValidation=true`, nothing else | ✅ bypassed |
| nothing configured (default) | ❌ fails with updated message + envFrom hint |
| unrelated `env.SOME_OTHER=x` | ❌ still fails (no false-positive) |
| `rrGitServer.enabled=false` | ✅ no-op |

Both values.yaml copies stay in sync; `helm lint` passes.

### DIFF
--- a/charts/retool/templates/_helpers.tpl
+++ b/charts/retool/templates/_helpers.tpl
@@ -684,13 +684,18 @@ Set MCP server service name
 {{/*
 Validate that exactly one blob-storage provider is configured when rrGitServer
 is enabled. Skipped when the user has plumbed the RR_BLOB_STORAGE_PROVIDER /
-RR_DEFAULT_*_* env vars in directly via environmentVariables/environmentSecrets,
+RR_DEFAULT_*_* env vars in directly via env/environmentVariables/environmentSecrets,
 which is treated as an opt-out from the first-class blobStorage config.
 No-op when rrGitServer is disabled.
 */}}
 {{- define "retool.rrGitServer.validateBlobStorage" -}}
 {{- if .Values.rrGitServer.enabled -}}
 {{- $hasDirectEnv := false -}}
+{{- range $name, $value := .Values.env -}}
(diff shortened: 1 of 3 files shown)


ryanartecona (approved):
_technically_ env vars can also come from `envFrom`, i.e. if you have a secret or configmap that you just want to splat all the keys/values in there into env vars on the backend pods. I don't think we can detect at template time at all in that case, but it might be prudent to also add an explicit disable for this sanity check in that case where it needs to be overridden.

JatinNanda:
> _technically_ env vars can also come from `envFrom`, i.e. if you have a secret or configmap that you just want to splat all the keys/values in there into env vars on the backend pods. I don't think we can detect at template time at all in that case, but it might be prudent to also add an explicit disable for this sanity check in that case where it needs to be overridden.

good call, added a skip

PR #308, opened 2026-06-08
agent-sandbox: validate required secrets + existing-secret DSN ref

JatinNanda:
Part of the **r2-cleanup** integration branch (merges into `r2-cleanup`, then `r2` via #303).

## Problems
1. **Silent empty postgres URL.** `{{ $as.postgres.url | default "" | b64enc }}` base64-encodes an empty value, so a deploy missing the DSN installs cleanly and the controller/proxy crash-loop at runtime.
2. **Unguarded required secrets.** `jwtPublicKey` (controller + proxy `process.exit(1)` without it) and `jwtPrivateKey` (the backend signs sandbox tokens with it) had no install-time check.
3. **Rigid / redundant postgres config.** Only a plaintext DSN was accepted, and enabling the sandbox against the *existing* database forced operators to re-enter host/user/password even though it just needs a separate schema in the same DB.

## Postgres connection — inherits the backend by default
The agent sandbox lives in the **same database** as the backend (separate schema, `agent_executor`). So **by default (nothing set under `agentSandbox.postgres`), the chart reuses the backend's connection** — host/port/database/user from `config.postgresql` (or the postgresql subchart), and `PGPASSWORD` from the same secret the backend uses. Enabling r2 on an existing deployment needs **no** new Postgres values.

To point the sandbox elsewhere, set exactly one (resolution order):
1. `postgres.url` — plaintext DSN.
2. `postgres.host` (+ `user` + `database`) — chart assembles `postgres://user@host:port/database`; password via `PGPASSWORD` from `postgres.password` or `postgres.passwordSecretName` (reuse a password-only secret).
3. `postgres.urlSecretName`/`urlSecretKey` — existing secret with the full DSN.
4. `externalSecret.name` — catch-all secret, `postgres-url` key.
5. *(default)* inherit `config.postgresql`.

**Edge gated by validateSecrets:** if the backend password comes via external secrets (`envFrom`) with no discrete key, inheritance can't reference it, so install fails with guidance to set an explicit option.

## App constraint & PGPASSWORD
The app passes a single connection string to Knex/`pg` (no split-field path). To carry a secret-sourced password without URL escaping, the chart omits the password from the DSN and supplies `PGPASSWORD`; node-postgres reads it when the connection string has no password (verified on `pg-connection-string` 2.4.0 + `pg` 8.11.3). `PGPASSWORD` is process-global but safe — the controller/proxy open exactly one Postgres connection.

## user/database characters (review follow-up)
`user`/`database` are embedded verbatim. Tested `pg-connection-string` 2.4.0: percent-encoding does **not** round-trip (userinfo is decoded *before* the `:` split; the path uses `decodeURI`), and Azure-style `user@servername` already parses correctly (splits on the **last** `@`). So `validateSecrets` rejects only the chars that genuinely break parsing — `:` `/` `?` `#`/whitespace in `user`, `?` `#`/whitespace in `database` — and points to options 1/3 otherwise. `@` is allowed.

## Other changes
- **`retool.agentSandbox.validateSecrets`** — fails on missing Postgres source (when inheritance is also unavailable), missing user/database on the assemble path, unsafe chars, or missing JWT public/private key.
- Controller/proxy URL block promoted to **`retool.agentSandbox.postgresUrlEnv`**.
- Chart-managed secret writes `postgres-url` only when a plaintext url is set.
- Docs in `values.yaml` lead with the inherit-by-default behavior.

## Audit of the other r2 workloads
- **mcp**: already fails when `oauthIntrospectionAuthToken` is absent. No change.
- **js_executor**: no secrets (none after #304). No change.

## Verification (`helm template`, agentSandbox.enabled=true)
| Case | Result |
|------|--------|
| subchart, `agentSandbox.postgres` unset | ✅ inherits `r2-postgresql` host + `postgres-password` |
| `config.postgresql` (plaintext pw), unset | ✅ inherits; `PGPASSWORD` from `postgresql-password`; port defaults 5432 |
| `config.postgresql.passwordSecretName`, unset | ✅ inherits; `PGPASSWORD` from that secret |
| external-secrets mode, unset | ❌ fails with guidance |
| `postgres.url` / `urlSecretName` / `externalSecret` | ✅ overrides inheritance |
| assemble + `passwordSecretName` / plaintext (incl. `p@ss/w:rd`) | ✅ `PGPASSWORD`; URL omits password |
| assemble `user=admin@pg` (Azure) | ✅ `postgres://admin@pg@host:5432/db` |
| assemble `user` `:`/`/`, `database` `?`/`#` | ❌ fails with guidance |
| missing jwt public / jwt private | ❌ fails (distinct messages) |
| `agentSandbox.enabled=false` | ✅ no-op |

Both values.yaml copies stay in sync; `helm lint` passes; no CI fixture enables agentSandbox.

### DIFF
--- a/charts/retool/templates/_helpers.tpl
+++ b/charts/retool/templates/_helpers.tpl
@@ -624,6 +624,100 @@ telemetry.retool.com/service-name: agent-sandbox-proxy
 {{- end -}}
 
 {{/*
+Validate that an enabled agent sandbox has its required secrets supplied. The
+controller and proxy fail to boot without a Postgres connection and a JWT
+public key, and the Retool backend needs the JWT private key to sign sandbox
+tokens. Each may come from a plaintext value, the per-key existing-secret refs,
+or the catch-all externalSecret.name. No-op when agentSandbox is disabled.
+*/}}
+{{- define "retool.agentSandbox.validateSecrets" -}}
+{{- if .Values.agentSandbox.enabled -}}
+{{- $as := .Values.agentSandbox -}}
+{{- $ext := $as.externalSecret.name -}}
+{{- if not (or $as.postgres.url $as.postgres.urlSecretName $as.postgres.host $ext) -}}
+{{- fail "agentSandbox.enabled requires a Postgres connection. Set one of: agentSandbox.postgres.url (DSN), agentSandbox.postgres.host + user + database (the chart assembles the DSN, password from postgres.password or postgres.passwordSecretName), agentSandbox.postgres.urlSecretName (existing secret holding the DSN), or agentSandbox.externalSecret.name." -}}
+{{- end -}}
+{{- if $as.postgres.host -}}
+{{- if not (and $as.postgres.user $as.postgres.database) -}}
+{{- fail "agentSandbox.postgres.host is set, so postgres.user and postgres.database are also required to assemble the DSN." -}}
+{{- end -}}
+{{- /*
+  user and database are embedded verbatim in the assembled DSN, so reject the
+  characters that would break URL parsing. '@' is allowed in user (managed
+  services like Azure use user@servername; the parser splits on the last '@'),
+  but ':' '/' and whitespace would be mis-parsed as a password/host/path. For
+  values needing other characters, supply a full DSN via postgres.url or
+  postgres.urlSecretName instead.
+*/}}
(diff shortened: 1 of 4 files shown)


ryanartecona (approved):
good!

PR #310, opened 2026-06-08
[fix[R2] Increase the AE proxy timeout to be inline with fix in retool-k8s

jamie-retool:
This brings a fix that we made over in retool-k8s to be over here too: https://app.graphite.com/github/pr/tryretool/retool-k8s/18000

### DIFF
--- a/charts/retool/values.yaml
+++ b/charts/retool/values.yaml
@@ -1071,7 +1071,7 @@ agentSandbox:
     # Defaults to http://<fullname>:3000 (same-cluster backend service).
     backendUrl: ''
     backendDomainSuffixes: ''
-    sandboxProxyTimeoutMs: ''
+    sandboxProxyTimeoutMs: 180000 # 3 minutes
     service:
       # Set to LoadBalancer or NodePort to expose the proxy externally.
       type: ClusterIP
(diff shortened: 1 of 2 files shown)


charts/retool/values.yaml
```diff
@@ -1071,7 +1071,7 @@ agentSandbox:
     # Defaults to http://<fullname>:3000 (same-cluster backend service).
     backendUrl: ''
     backendDomainSuffixes: ''
-    sandboxProxyTimeoutMs: ''
+    sandboxProxyTimeoutMs: 180000 # 3 minutes
```
ryanartecona: these map to env vars so should stay strings ideally

```suggestion
    sandboxProxyTimeoutMs: '180000' # 3 minutes
```

values.yaml
```diff
@@ -1071,7 +1071,7 @@ agentSandbox:
     # Defaults to http://<fullname>:3000 (same-cluster backend service).
     backendUrl: ''
     backendDomainSuffixes: ''
-    sandboxProxyTimeoutMs: ''
+    sandboxProxyTimeoutMs: 180000 # 3 minutes
```
ryanartecona:
```suggestion
    sandboxProxyTimeoutMs: '180000' # 3 minutes
```

ryanartecona (approved):

PR #315, opened 2026-06-09
r2-cleanup → r2: R2 enablement polish (master switch, same-origin proxy, secrets, git-server split)

JatinNanda:
## Summary

Merges the `r2-cleanup` integration branch into the `r2` preview branch. This batches up the polish work that meaningfully simplifies how R² is enabled on existing self-hosted deployments.

### Headline changes
- **Single `r2.enabled` master switch** (#313) — one toggle turns on `r2Agent`, `jsExecutor`, `agentSandbox`, and `mcp`. Each component's `enabled` defaults to `null` and inherits the switch; set it explicitly to override.
- **Same-origin agent-sandbox proxy / no extra ingress** (#302) — leaving `agentSandbox.frontendWsProxyDomain` empty makes the backend serve the sandbox same-origin via the main ingress. No dedicated `agent-proxy.*` domain, ingress, or extra HTTPRoute required; the catch-all ingress is untouched.
- **Agent-sandbox secrets + Postgres** (#308) — only the JWT keypair is required; `encryptionKey`/`apiSecret` are optional. Postgres is inherited from the backend by default (separate `agent_executor` schema), with explicit DSN / fields / existing-secret options for the cases where it can't be inherited. Adds required-secret validation and existing-secret DSN refs.
- **Optional split git server** (#309) — `rrGitServer.separate.enabled` runs the git server as its own Deployment + Service; backend proxies to it and MCP auto-points at it.
- **rrGitServer blob-storage env** (#307) — accepts blob-storage env vars from `.Values.env`, plus a `skipBlobStorageValidation` escape hatch for `envFrom`-injected vars.
- **js-executor** (#304) — drops backend-shared env inheritance (own `env`/`environmentSecrets`/`environmentVariables`) and resizes resources (equal memory request/limit so JSE's 80%-of-limit rejection works correctly).

### Included commits
- `52a1d01` [feat][r2] add single r2.enabled master switch for R2 components (#313)
- `b049145` ci: test coverage for r2 workloads (js-executor, agent-sandbox, r2Agent) (#312)
- `480fb81` agent-sandbox: validate required secrets + existing-secret DSN ref (#308)
- `8b0d492` [feat][r2] optionally split rrGitServer into its own deployment (#309)
- `7675ccf` re-sync values.yaml with chart copy after #302
- `50c91c0` Document self-hosted same-origin agent-sandbox proxy (no extra ingress) (#302)
- `b88cfdd` rrGitServer: accept blob-storage env vars from .Values.env (#307)
- `015e98f` js-executor: drop backend-shared env inheritance + resize resources (#304)

### Notes
- Values changes are mirrored in both `charts/retool/values.yaml` and the root `values.yaml` per chart convention.
- New CI test fixtures cover the enable options (r2-enabled, per-component enables, agent-sandbox secret/postgres variants, rr-git-server separate).

### DIFF
--- a/charts/retool/ci/test-agent-sandbox-enabled-option.yaml
+++ b/charts/retool/ci/test-agent-sandbox-enabled-option.yaml
@@ -1,0 +1,77 @@
+# Agent Sandbox — external secret + dedicated proxy domain WITH ingress.
+#
+# This is the "max surface" scenario: controller + proxy deployments, the
+# job-template ConfigMap, RBAC, headless/proxy services, proxy ingress + TLS,
+# the image-prepuller + seccomp DaemonSets, the smarter-device-manager device
+# plugin DaemonSet, the NetworkPolicies, and both PDBs.
+#
+# Secret/Postgres sourcing here uses externalSecret.name (Postgres OPTION 4:
+# the secret's postgres-url key). The other secret/Postgres precedence paths and
+# the same-origin (no-ingress) proxy mode are covered by sibling files:
+#   - test-agent-sandbox-inline-secrets-option.yaml      (inline secrets, plaintext DSN, same-origin/no ingress, hostPath tun)
+#   - test-agent-sandbox-postgres-fields-option.yaml     (assemble DSN from fields + PGPASSWORD secret)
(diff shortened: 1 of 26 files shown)


ryanartecona (approved):
WOW HUGE

PR #322, opened 2026-06-11
[agent-sandbox] expose sandboxReadyTimeoutMs -> SANDBOX_READY_TIMEOUT_MS

JatinNanda:
## What
Adds `agentSandbox.sandbox.sandboxReadyTimeoutMs` (default `20000`) and emits `SANDBOX_READY_TIMEOUT_MS` in the agent-sandbox job-template env, alongside `SANDBOX_IDLE_TIMEOUT_MS` / `SANDBOX_GLOBAL_LIFETIME_MS`.

## Why
The agent-executor made the sandbox connect timeout env-configurable (`config.ts` `readyTimeoutMs` ← `SANDBOX_READY_TIMEOUT_MS`, in retool_development #79287/#79443), but the chart never set it — so the job-template used the image default of 20s. Interactive sandbox boot (gVisor + bundle load) can take ~20–25s, surfacing `agent sandbox <id> did not connect within 20000ms`. Until now the only workaround was a manual `readytimeout-patch` initContainer that sed-ed the bundle — this replaces that with a first-class knob.

## Default
`20000`, matching the code default (no behavior change on upgrade). Operators who need more set e.g. `45000`.

## Notes
- Both `values.yaml` and `charts/retool/values.yaml` updated identically (sync gate).
- Verified `helm template` renders `SANDBOX_READY_TIMEOUT_MS` in the job-template.
- Internal-onprem balloons need a separate retool-k8s values override to actually raise it (chart default stays conservative).


slaminad (approved):
goat

PR #324, opened 2026-06-11
Merge r2 → main: R² (React Retool) self-hosted support (chart 6.12.0)

JatinNanda:
## What

Lands the full **R² (React Retool)** self-hosted feature set from the long-lived `r2` branch onto the published `main` chart line. Chart version **6.10.6 → 6.12.0**.

This is a real merge commit (not a squash) so the **33-commit R² history is preserved** — see the Commits tab.

### What's included
- **`rr.*` values layout** (#321) — one top-level `rr.enabled` master switch over `rr.jsExecutor` / `rr.agent` / `rr.agentSandbox`, with null-inherit per-component overrides. Fail-loud guard (`retool.rr.validateLegacyValues`) on the old `r2.*` / top-level keys.
- **js-executor** workload (#275), self-contained env, Guaranteed QoS.
- **rr.agent** server-side agent worker.
- **agent-sandbox** service (#281, #290, #308): controller + proxy + ephemeral Job sandboxes, flexible Postgres sourcing (inherit backend by default), flexible secret sourcing (JWT required, encryption/api optional), same-origin proxy by default (no extra ingress), NetworkPolicy, device plugin.
- **git server** (#296) in-process by default; optional split deployment (#309).
- **shared blob storage** block (s3/gcs/azure) + raw-env escape hatch.
- **MCP server** (#285), top-level and opt-in (not part of the master switch, #316).

## Conflict resolution
- `charts/retool/Chart.yaml` **version → 6.12.0** (the R² release; subsumes main's 6.10.6 from #305).
- main's #305 `startupProbe.successThreshold` example and r2's temporal dependency-condition change are both preserved (auto-merged).

## Prerequisites (done)
Consumer migrations to the (non-backwards-compatible) `rr.*` layout landed **before** this merge:
- retool-k8s **#18177** — internal-onprem + admin + managed-self-hosted
- retool-k8s **#18178** — nuon

## Validation (on the merged tree)
- `helm lint` clean
- All **19** `ci/*option.yaml` permutations render
- Legacy-values guard fires on stale `r2.*` keys

## Not in this PR
- **Publishing** to charts.retool.com (6.12.0) — deliberate separate rollout step.
- **retool-self-hosted-blueprints** rr.* migration — still to do before that repo consumes the published chart (it currently pins `ref=r2`, so unaffected by this merge).

> ⚠️ **Merge note:** repo is configured squash-only, which would flatten the 33 commits. To preserve history on `main`, enable merge commits before merging this PR.


charts/retool/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 6.10.6
+version: 6.12.0
```
ryanartecona, line 5: should be 6.11.0 no?

charts/retool/Chart.yaml
```diff
@@ -13,4 +13,4 @@ dependencies:
     condition: postgresql.enabled
   - name: retool-temporal-services-helm
     version: 1.1.5
-    condition: retool-temporal-services-helm.enabled,workflows.enabled
+    condition: retool-temporal-services-helm.enabled
```
ryanartecona, line 16: this is a breaking change, which we were otherwise able to avoid. I think @lukefoster11 you may have introduced this IIRC? is this load bearing for r2 stuff or can we undo this part?

PR #326, opened 2026-06-11
Merge r2 → main: R² (React Retool) self-hosted support (chart 6.11.0)

JatinNanda:
Merges the full **R² (React Retool)** self-hosted feature set from the long-lived `r2` branch into `main`. Opening `r2` directly so all 33 commits are preserved in history.

## What's included
- **`rr.*` values layout** (#321) — top-level `rr.enabled` master switch over `rr.jsExecutor` / `rr.agent` / `rr.agentSandbox`, with null-inherit per-component overrides + fail-loud guard on the old `r2.*` keys.
- **js-executor** (#275), **rr.agent** worker, **agent-sandbox** (controller/proxy/ephemeral Job sandboxes, flexible Postgres + secrets, same-origin proxy by default), **git server** (in-process + optional split #309), **shared blob storage** (s3/gcs/azure), **MCP server** (#285, top-level opt-in).
- **jwtPublicKey JSON fix** (#325) for multi-line PEM keys in the sandbox job-template.

## Conflict
- `charts/retool/Chart.yaml` version (main 6.10.6 vs r2 6.12.0) → resolve to **6.12.0**. main's #305 `startupProbe.successThreshold` (5→1) should be kept.

## Prerequisites (done)
Consumer migrations to the `rr.*` layout landed first: retool-k8s **#18177** (internal-onprem + admin + MSH) and **#18178** (nuon).

## Notes
- Chart stays a **minor** release (6.12.0): additive for published consumers, all R² switches default off; the rename guard can't fire on the published line (those keys never existed there).
- Repo is currently squash-only — enable merge commits before merging if history on `main` is to be preserved.
- Supersedes #324 (the `jatin/r2-to-main` intermediate-branch version).


ryanartecona (approved):
🚀 🚀 🚀 🚀 🚀

PR #330, opened 2026-06-12
[codex] Set agent sandbox proxy URL default

arnold-retool:
## Summary
- Defaults the Agent Sandbox proxy URL to the namespace-qualified in-cluster Service DNS name for server-side callers.
- Exposes that URL through AGENT_SANDBOX_PROXY_INGRESS_DOMAIN for backend/workflow callers and MCP, without adding backend-only secrets to MCP.
- Keeps frontend WebSocket proxy configuration separate and documents the new same-cluster default, with rr.agentSandbox.proxyUrl available for explicit overrides.

## Validation
- helm template for Agent Sandbox enabled and disabled cases
- helm template for MCP + Agent Sandbox and explicit proxyUrl override cases
- helm lint charts/retool with test-install-values.yaml

### DIFF
--- a/charts/retool/templates/_helpers.tpl
+++ b/charts/retool/templates/_helpers.tpl
@@ -646,6 +646,26 @@ telemetry.retool.com/service-name: agent-sandbox-proxy
 {{- end -}}
 
 {{/*
+Internal URL for server-side callers to reach the agent sandbox proxy Service.
+The fully-qualified cluster DNS name keeps this default valid even when callers
+run outside the proxy Service's namespace.
+*/}}
+{{- define "retool.agentSandbox.proxyUrl" -}}
+{{- .Values.rr.agentSandbox.proxyUrl | default (printf "http://%s.%s.svc.cluster.local:%s" (include "retool.agentSandbox.proxy.name" .) .Release.Namespace (toString .Values.rr.agentSandbox.proxy.port)) -}}
+{{- end -}}
+
+{{/*
+Agent sandbox proxy env var for server-side callers. Keep this separate from
+AGENT_*_FRONTEND_WS_PROXY_DOMAIN, which is browser-facing public config.
+*/}}
+{{- define "retool.agentSandbox.proxyEnvVars" -}}
+{{- if eq (include "retool.rr.componentEnabled" (dict "root" . "component" "agentSandbox")) "1" }}
+- name: AGENT_SANDBOX_PROXY_INGRESS_DOMAIN
+  value: {{ include "retool.agentSandbox.proxyUrl" . | quote }}
+{{- end }}
+{{- end -}}
+
+{{/*
 Validate that an enabled agent sandbox has its required secrets supplied. The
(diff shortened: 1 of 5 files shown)


charts/retool/templates/_helpers.tpl
```diff
@@ -645,6 +645,28 @@ app.kubernetes.io/component: proxy
...
+
+{{/*
+Agent sandbox proxy env vars for server-side callers. Keep this separate from
+AGENT_*_FRONTEND_WS_PROXY_DOMAIN, which is browser-facing public config.
+*/}}
+{{- define "retool.agentSandbox.proxyEnvVars" -}}
+{{- if eq (include "retool.rr.componentEnabled" (dict "root" . "component" "agentSandbox")) "1" }}
+- name: AGENT_EXECUTOR_PROXY_INGRESS_DOMAIN
+  value: {{ include "retool.agentSandbox.proxyUrl" . | quote }}
+- name: AGENT_SANDBOX_PROXY_INGRESS_DOMAIN
```
westrik, line 663: `AGENT_EXECUTOR_PROXY_INGRESS_DOMAIN` is an alias for `AGENT_SANDBOX_PROXY_INGRESS_DOMAIN`. We should continue only defining the latter to keep naming consistent.

charts/retool/templates/_helpers.tpl
```diff
@@ -645,6 +645,28 @@ app.kubernetes.io/component: proxy
 telemetry.retool.com/service-name: agent-sandbox-proxy
 {{- end -}}
 
+{{/*
+Internal URL for server-side callers to reach the agent sandbox proxy Service.
+The fully-qualified cluster DNS name keeps this default valid even when callers
+run outside the proxy Service's namespace.
+*/}}
+{{- define "retool.agentSandbox.proxyUrl" -}}
+{{- .Values.rr.agentSandbox.proxyUrl | default (printf "http://%s.%s.svc.cluster.local:%s" (include "retool.agentSandbox.proxy.name" .) .Release.Namespace (toString .Values.rr.agentSandbox.proxy.port)) -}}
```
westrik, line 654: just curious — do we actually need the FQDN here?

arnold-retool: we want an unambiguous way to contact the AE proxy, even if they are different pods (jailed vs non jailed)

charts/retool/templates/_helpers.tpl
arnold-retool: thanks

charts/retool/templates/_helpers.tpl
arnold-retool: We have an issue on admin deploy where the MCP server cannot contact sandboxes because requests are being routed outside the VPC and coming back in

arnold-retool: FWIW on qa we have a weird elastic load balancing URL. if this isn't possible to populate with a good default then we'll leave it blank for now? @westrik

JatinNanda (approved):

PR #331, opened 2026-06-12
fix(rr): decouple agentSandbox postgres from externalSecret (use postgres.urlSecretName); require encryption key

JatinNanda:
## what

setting `rr.agentSandbox.externalSecret.name` (intended just for the JWT keypair) silently overloaded two unrelated things, breaking enablement on a deployment that only wanted to reuse its existing backend postgres:

- **postgres hijack** — `postgresUrlEnv` routed `AGENT_SANDBOX_POSTGRES_URL` to the external secret's `postgres-url` key. missing key → `getaddrinfo ENOTFOUND` at runtime. the inherit-the-backend path was unreachable whenever `externalSecret.name` was set, and `validateSecrets` counted it as an explicit pg source so nothing failed at template time.
- **encryption key was treated as optional** — but it isn't. the proxy derives the sandbox-iframe asset-token HMAC key from `AGENT_SANDBOX_ENCRYPTION_KEY` (`agent_executor/proxy/src/config.ts` `getAssetTokenHmacSecret`) and **throws the first time it serves a sandbox** if it's missing. the pod comes up green and then fails at first use.

separately, the snapshot blob-storage env never reached the **workflow worker**, even though the agent/snapshot temporal activities run there.

## changes

- **`externalSecret.name` now covers only the JWT/encryption keys — it never sources postgres.** to read a DSN from that same secret, point `postgres.urlSecretName` at it (Option 3; `urlSecretKey` defaults to `postgres-url`). otherwise postgres inherits `config.postgresql` — an existing deployment needs no DB config.
- **require** the agent sandbox encryption key in `validateSecrets` (mirrors the JWT-key checks) — fail loudly at render instead of shipping a pod that breaks at first sandbox use; both fail hints point at `postgres.urlSecretName`
- **render `gitServer.commonEnv` (`RR_DEFAULT_*`) onto the workflow worker** when `rr.gitServer.enabled`. the `agentExecutor`/`snapshotRetention` temporal activities run on `WORKFLOW_TEMPORAL_WORKER` and read snapshot blob storage (`getBlobStoreForSnapshots`, `RR_SNAPSHOTS_*` → `RR_DEFAULT_*` fallback), but `commonEnv` was only injected onto the backend + standalone git-server pod — so the worker had no creds and snapshot access failed there.
- `values.yaml` (both copies): drop `postgres-url` from the `externalSecret.name` key list, mark `encryptionKey` required (64 hex, `openssl rand -hex 32`)
- bump chart 6.11.1 → 6.11.2; update `ci/` fixtures to supply the now-required encryption key

## incorporates #332

pulled in @ryanartecona's `passwordSecretName` support so the two don't collide. in the final shape it lives in the `postgres.urlSecretName` branch: the blueprints case (DSN from the agent-sandbox secret, password from the auto-rotated main secret) is `postgres.urlSecretName: agent-sandbox-env` + `postgres.passwordSecretName: retool-config`.

> note: an earlier revision used a `postgres.fromExternalSecret` boolean — redundant with `postgres.urlSecretName`, since removed in favor of it.

## testing

`helm template` verified: `urlSecretName` routes to the secret's `postgres-url` (+ `PGPASSWORD` when `passwordSecretName` set); JWT-only `externalSecret` inherits the backend DSN; render **fails** with a clear error when no encryption-key source is set; the workflow worker now renders `RR_DEFAULT_*` when `gitServer.enabled` + `blobStorage` is set. all `ci/` fixtures render; `helm lint` clean.

note: same-origin-on-HTTPS (a separate reported issue) is a backend gap, not chart-fixable, and is out of scope here.

### DIFF
--- a/charts/retool/Chart.yaml
+++ b/charts/retool/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 6.11.2
+version: 6.11.3
 maintainers:
   - name: Retool Engineering
     email: engineering+helm@retool.com
--- a/charts/retool/ci/test-agent-sandbox-enabled-option.yaml
+++ b/charts/retool/ci/test-agent-sandbox-enabled-option.yaml
@@ -31,6 +31,9 @@ rr:
       name: agent-sandbox-secrets
 
     postgres:
+      # Option 4 is opt-in: route Postgres to the external secret's postgres-url key.
+      # externalSecret.name alone only covers the JWT/encryption keys, not Postgres.
+      fromExternalSecret: true
       schema: agent_executor
       poolMax: 10
 
--- a/charts/retool/ci/test-agent-sandbox-enabled-option.yaml
+++ b/charts/retool/ci/test-agent-sandbox-enabled-option.yaml
@@ -7,9 +7,10 @@ rr:
   # the image-prepuller + seccomp DaemonSets, the smarter-device-manager device
   # plugin DaemonSet, the NetworkPolicies, and both PDBs.
   #
-  # Secret/Postgres sourcing here uses externalSecret.name (Postgres OPTION 4:
-  # the secret's postgres-url key). The other secret/Postgres precedence paths and
-  # the same-origin (no-ingress) proxy mode are covered by sibling files:
+  # Secrets come from externalSecret.name (JWT/encryption); Postgres reads its DSN
+  # from that same Secret via postgres.urlSecretName (Option 3). The other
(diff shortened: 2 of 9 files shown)


ryanartecona:
uhhh what do you think about instead of `rr.agentSandbox.postgres.fromExternalSecret: true`, we just mirror the pattern we have for other stuff already and introduce `rr.agentSandbox.postgres.urlSecret{Name,Key}`? that is IMO more flexible & straightforward with the slight tradeoff of being slightly more verbose.

lukefoster11 (approved):

PR #334, opened 2026-06-15
fix(rr): inherit backend SSL for agent sandbox postgres (sslmode=no-verify)

JatinNanda:
## what

When `rr.agentSandbox.postgres` is left to **inherit** the backend connection (`config.postgresql`), the chart assembled `AGENT_SANDBOX_POSTGRES_URL` with **no `sslmode`**. The agent-sandbox controller/proxy connect via Knex/`pg`, which only does TLS if the connection string says so — so an SSL-required RDS rejects them:

```
no pg_hba.conf entry for host "...", user "retool_internal_user", database "hammerhead_admin_prod", no encryption
```

The backend connects to the same DB fine because it applies SSL (`common/resources/connectionStringUtil.ts` → `sslmode=no-verify`); the inherited sandbox DSN didn't carry it.

## fix

Append `?sslmode=no-verify` to the inherited DSN when the backend uses SSL (`retool.postgresql.ssl_enabled` is true), mirroring the backend. No effect when SSL is off (e.g. the in-cluster postgresql subchart). The explicit `postgres.url` / `urlSecretName` paths are unchanged — operators put `sslmode` in those themselves.

## testing

- `helm template`: external pg + `ssl_enabled: true` → `…/hammerhead_admin_prod?sslmode=no-verify`; in-cluster subchart → no `sslmode`. All `ci/` fixtures render; `helm lint` clean.
- **New CI fixture** `ci/test-agent-sandbox-inherit-ssl-option.yaml` exercises the external-SSL inherit path (`postgresql.enabled: false` + `config.postgresql.ssl_enabled: true`), which no existing fixture covered — locks in the `sslmode` suffix against future regressions.
- Chart bumped 6.11.3 → 6.11.4.

### DIFF
--- a/charts/retool/Chart.yaml
+++ b/charts/retool/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 6.11.3
+version: 6.11.4
 maintainers:
   - name: Retool Engineering
     email: engineering+helm@retool.com
--- a/charts/retool/ci/test-agent-sandbox-inherit-ssl-option.yaml
+++ b/charts/retool/ci/test-agent-sandbox-inherit-ssl-option.yaml
@@ -1,0 +1,38 @@
+rr:
+
+  # Agent Sandbox — inherit an EXTERNAL, SSL-required Postgres. Complements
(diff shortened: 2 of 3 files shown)


lukefoster11 (approved):
"%s %s %s %s %s" ("this") ("looks") ("good") ("to") ("me")

PR #336, opened 2026-06-17
Allow configuring host/scheme on liveness, readiness, and startup probes (fixes code-executor on IPv6 clusters)

### LINKED ISSUE #335: code-executor CrashLoopBackOff on IPv6-primary clusters — httpGet probes target the Pod IPv6 address but the service listens on IPv4 only
### Summary

On an IPv6-primary (or single-stack IPv6) cluster, the `code-executor` pod never becomes
ready and ends up in `CrashLoopBackOff`. The liveness/readiness probes fail with
`connection refused`. The container itself starts fine — the problem is that the
`code-executor` HTTP server on `:3004` binds to IPv4 only, while the kubelet probes the
Pod's primary address, which is IPv6.

### Environment

- Chart: `retool` 6.11.4 (probe template is identical on `main`)
- `codeExecutor.enabled: true`, `workflows.enabled: true`
- Kubernetes: EKS, **IPv6-primary** cluster (Pods get an IPv6 primary IP)
- Image: `tryretool/code-executor-service` (tag tracks `image.tag`)

### What happens

```
$ kubectl get pod -l app.kubernetes.io/name=retool-code-executor
NAME                                  READY   STATUS             RESTARTS   AGE
retool-code-executor-7b9c8f9c-abcde   0/1     CrashLoopBackOff   6          7m

$ kubectl describe pod retool-code-executor-7b9c8f9c-abcde
...
  Warning  Unhealthy  ...  Liveness probe failed: Get
    "http://[2600:1f13:xxxx:xxxx::a]:3004/api/checkHealth":
    dial tcp [2600:1f13:xxxx:xxxx::a]:3004: connect: connection refused
  Warning  Unhealthy  ...  Readiness probe failed: Get
    "http://[2600:1f13:xxxx:xxxx::a]:3004/api/checkHealth":
    dial tcp [2600:1f13:xxxx:xxxx::a]:3004: connect: connection refused
```

### Root cause

The probe (from `templates/deployment_code_executor.yaml`) sets no `host`, so the kubelet
uses the Pod IP — which is IPv6 on this cluster:

```yaml
        livenessProbe:
          httpGet:
            path: {{ .Values.livenessProbe.path }}
            port: 3004
```

But the server only listens on IPv4. Verified from inside the pod:

```
# IPv4 loopback — works
$ kubectl exec ... -- node -e "require('http').get('http://127.0.0.1:3004/api/checkHealth',r=>console.log(r.statusCode))"
200

# IPv6 loopback / Pod IPv6 — refused
$ kubectl exec ... -- node -e "require('http').get('http://[::1]:3004/api/checkHealth').on('error',e=>console.log(e.code))"
ECONNREFUSED

$ kubectl exec ... -- sh -c 'ss -tlnp | grep 3004'
LISTEN 0  511   0.0.0.0:3004   0.0.0.0:*      # IPv4 only; no :::3004
```

### Why this can't be worked around with the chart today

`livenessProbe`/`readinessProbe` only expose `enabled`, `path`, and timing fields — there is
no `host` or `scheme`. So there is no supported way to point the probe at the IPv4 loopback
without forking the chart or post-rendering the manifests.

### Proposed fix

Add optional `host` (and `scheme`) passthrough to the httpGet probes. Operators on
IPv6/dual-stack clusters can then set `livenessProbe.host: 127.0.0.1` /
`readinessProbe.host: 127.0.0.1`, which makes the kubelet probe the IPv4 loopback the
server is already listening on. Fully backwards-compatible (the fields are omitted when
unset, so default rendering is unchanged). PR to follow.

> Deeper fix (out of scope for this chart): have `code-executor-service` listen on `::`
> (dual-stack) so it answers on the Pod IP directly.

carlzogh-ai:
### What

Adds optional `host` and `scheme` fields to the `livenessProbe`, `readinessProbe`, and
`startupProbe` httpGet probes, wired into every deployment that renders from those shared
values (`backend`, `code-executor`, `dbconnector`, `workflows`, `js-executor`, and the
workers in `_workers.tpl`).

### Why

On IPv6-primary / dual-stack clusters, the `code-executor` pod CrashLoops because the
kubelet probes the Pod's primary IP (IPv6) but the service listens on IPv4 only, so the
probe fails with `connection refused`. There is currently no supported way to point the
probe at the IPv4 loopback — the probe values only expose `enabled`, `path`, and timing
fields. See #335 for the full reproduction and evidence.

With this change, operators on such clusters can set:

```yaml
livenessProbe:
  host: 127.0.0.1
readinessProbe:
  host: 127.0.0.1
```

which makes the kubelet probe the IPv4 loopback the server is already bound to. This is the
standard Kubernetes mechanism (`httpGet.host`) for exactly this situation.

### Backwards compatibility

Fully backwards-compatible. Both fields are rendered only `{{- with ... }}`, so when unset
(the default) the manifests are byte-for-byte identical to before.

Verified with `helm template`:

- Defaults → no `host`/`scheme` emitted anywhere.
- `--set livenessProbe.host=127.0.0.1 --set readinessProbe.host=127.0.0.1` → `host` rendered
  inside `httpGet` across all affected components.

### Notes

The deeper root cause (the `code-executor-service` image binding IPv4-only instead of `::`)
is out of scope for this chart; this PR gives operators a supported workaround at the chart
level. Happy to add a chart version bump if the release process expects it in-PR.

Fixes #335

### DIFF
--- a/charts/retool/templates/_workers.tpl
+++ b/charts/retool/templates/_workers.tpl
@@ -358,6 +358,12 @@ spec:
           httpGet:
             path: {{ $.Values.livenessProbe.path }}
             port: {{ $healthcheckPort }}
+            {{- with $.Values.livenessProbe.host }}
+            host: {{ . | quote }}
+            {{- end }}
+            {{- with $.Values.livenessProbe.scheme }}
+            scheme: {{ . }}
+            {{- end }}
           initialDelaySeconds: {{ $.Values.livenessProbe.initialDelaySeconds }}
           timeoutSeconds: {{ $.Values.livenessProbe.timeoutSeconds }}
           failureThreshold:  {{ $.Values.livenessProbe.failureThreshold }}
@@ -367,6 +373,12 @@ spec:
           httpGet:
             path: {{ $.Values.readinessProbe.path }}
             port: {{ $healthcheckPort }}
+            {{- with $.Values.readinessProbe.host }}
+            host: {{ . | quote }}
+            {{- end }}
+            {{- with $.Values.readinessProbe.scheme }}
+            scheme: {{ . }}
+            {{- end }}
           initialDelaySeconds: {{ $.Values.readinessProbe.initialDelaySeconds }}
           timeoutSeconds: {{ $.Values.readinessProbe.timeoutSeconds }}
           successThreshold: {{ $.Values.readinessProbe.successThreshold }}
--- a/charts/retool/templates/deployment_agent_sandbox.yaml
+++ b/charts/retool/templates/deployment_agent_sandbox.yaml
@@ -366,6 +366,12 @@ spec:
             httpGet:
               path: /livez
               port: http
+              {{- with .Values.livenessProbe.host }}
+              host: {{ . | quote }}
+              {{- end }}
+              {{- with .Values.livenessProbe.scheme }}
+              scheme: {{ . }}
+              {{- end }}
             initialDelaySeconds: 10
             periodSeconds: 10
             timeoutSeconds: 5
@@ -374,6 +380,12 @@ spec:
             httpGet:
               path: /health
               port: http
+              {{- with .Values.readinessProbe.host }}
+              host: {{ . | quote }}
+              {{- end }}
+              {{- with .Values.readinessProbe.scheme }}
+              scheme: {{ . }}
+              {{- end }}
             initialDelaySeconds: 1
             periodSeconds: 2
             timeoutSeconds: 3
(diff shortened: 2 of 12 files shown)


carlzogh-ai:
Thanks @greptile-apps — good catch. Fixed in 598d9f2 by wiring the same `host`/`scheme` passthrough into the remaining HTTP probes that defined them inline and didn't consume the shared probe values: `deployment_jobs.yaml` (jobs-runner), `deployment_multiplayer_ws.yaml`, `deployment_mcp.yaml`, and both agent-sandbox Deployments in `deployment_agent_sandbox.yaml`.

Now a single global `livenessProbe.host`/`readinessProbe.host` reaches every HTTP-probed component, so the documented IPv6 workaround is no longer partial. Still rendered only `{{- with ... }}`, so defaults are byte-for-byte unchanged.

Verified with `helm template` (host set): `host: "127.0.0.1"` now renders for jobs-runner (2), multiplayer (2), MCP (2) and agent-sandbox (4), in addition to backend/code-executor/dbconnector/workflows/js-executor/workers; and with defaults it renders nowhere.

JatinNanda (approved):
lg2m, thanks!

PR #337, opened 2026-06-18
[fix][MCP] MCP Configuration improvements

arnold-retool:
## Summary

This PR improves the Helm chart setup path for self-hosted MCP deployments.

**Example shape**

```
mcp:
    enabled: true
    config:
      # Public Retool host that serves OAuth metadata and /api/oauth2/*.
      # Host-only is preferred; the chart also accepts https://example.retool.com.
      oauthMainDomain: example.retool.com

      # Secret-backed token used by MCP to call /api/oauth2/introspect.
      # oauthIntrospectionAuthTokenSecretKey defaults to oauthIntrospectionAuthToken.
      oauthIntrospectionAuthTokenSecretName: oauth-introspection-auth-token

      # Optional public Retool origin for client-facing MCP upload URLs.
      # Do not include /mcp.
      mcpServiceExternalUrl: https://example.retool.com

      # Optional MCP concurrency tuning. Renders MCP_MAX_TRANSPORT_SESSIONS.
      maxTransportSessions: 1000
```

**Some** **more** **details**

- Adds a dedicated backend API Service for MCP OAuth metadata routes, exposing the main backend API listener on port 3001.
- Adds a dedicated backend API Service and routes existing MCP backendApi Ingress/HTTPRoute entries to it
- Replaces old MCP RETOOL_URL env wiring with MCP_SERVICE_EXTERNAL_URL.
- Keeps retoolUrl as a deprecated alias for mcpServiceExternalUrl.
- Normalizes mcpServiceExternalUrl and oauthMainDomain.
- Requires oauthMainDomain and an OAuth introspection token source when mcp.enabled=true.
- Supports Secret-backed introspection token config without requiring plaintext values.
- Avoids duplicate env vars when OAUTH_MAIN_DOMAIN or OAUTH_INTROSPECTION_AUTH_TOKEN is supplied through mcp.environmentVariables.
- Documents optional MCP config, including maxTransportSessions.

## Validation

- helm lint charts/retool -f charts/retool/ci/test-install-values.yaml -f charts/retool/ci/test-mcp-enabled-option.yaml
- helm template for MCP deployment with CI values
- helm template for Secret-backed production-style MCP config
- helm template failure checks for missing oauthMainDomain and missing introspection token
- git diff --check
- Redeployed arniew-mcp balloon instance with the local chart
- Verified MCP connection through the connected arniew-mcp MCP server:
    - listed environments
    - listed apps
    - listed resources
    - listed folders

### DIFF
--- a/charts/retool/ci/test-mcp-agent-sandbox-secret-option.yaml
+++ b/charts/retool/ci/test-mcp-agent-sandbox-secret-option.yaml
@@ -1,6 +1,11 @@
+rr:
+  agentSandbox:
+    enabled: true
+    externalSecret:
+      name: agent-sandbox-jwt
+
 mcp:
   enabled: true
   config:
+    oauthMainDomain: example.retool.com
     oauthIntrospectionAuthToken: test-oauth-introspection-token
-    agentSandboxJwtPrivateKeySecretName: agent-sandbox-jwt
-    agentSandboxJwtPrivateKeySecretKey: private-key
--- a/charts/retool/templates/_helpers.tpl
+++ b/charts/retool/templates/_helpers.tpl
@@ -24,34 +24,18 @@ If release name contains chart name it will be used as a full name.
 {{- end }}
 
 {{/*
-Whether MCP routing needs the main Retool Service to expose the backend API
-listener in addition to the primary frontend-facing port.
+Append a suffix to retool.fullname while preserving the 63-character DNS label
+limit used by Service names and label values.
 */}}
-{{- define "retool.mcp.needsBackendApi" -}}
-{{- $mcp := .Values.mcp | default dict -}}
-{{- $mcpIngress := $mcp.ingress | default dict -}}
-{{- $mcpHttpRoute := $mcp.httpRoute | default dict -}}
-{{- $needsBackendApi := false -}}
-{{- if and .Values.ingress.enabled $mcp.enabled $mcpIngress.enabled -}}
-{{- range ($mcpIngress.paths | default list) -}}
-{{- if eq (.target | default "mcp") "backendApi" -}}
-{{- $needsBackendApi = true -}}
-{{- end -}}
-{{- end -}}
-{{- end -}}
-{{- if and .Values.httpRoute.enabled $mcp.enabled $mcpHttpRoute.enabled -}}
-{{- range ($mcpHttpRoute.rules | default list) -}}
-{{- if eq (.target | default "mcp") "backendApi" -}}
-{{- $needsBackendApi = true -}}
-{{- end -}}
-{{- end -}}
-{{- end -}}
-{{- if $needsBackendApi -}}true{{- else -}}false{{- end -}}
+{{- define "retool.fullnameWithSuffix" -}}
+{{- $root := index . 0 -}}
+{{- $suffix := index . 1 -}}
+{{- printf "%s-%s" (include "retool.fullname" $root | trunc (sub 62 (len $suffix) | int) | trimSuffix "-") $suffix | trunc 63 | trimSuffix "-" -}}
 {{- end }}
 
 {{/*
 Render an MCP-related Ingress path. By default paths route to the MCP service;
-target: backendApi routes to the main backend API listener instead.
+target: backendApi routes to the main backend API Service instead.
 */}}
 {{- define "retool.ingress.mcpPath" -}}
 {{- $root := .root -}}
@@ -65,7 +49,7 @@ target: backendApi routes to the main backend API listener instead.
 {{- $servicePort := $path.port | default ($mcpService.externalPort | default 4010) -}}
 {{- $pathType := $path.pathType | default "ImplementationSpecific" -}}
 {{- if eq $target "backendApi" -}}
-{{- $serviceName = include "retool.fullname" $root -}}
+{{- $serviceName = include "retool.backendApi.name" $root -}}
 {{- $servicePort = $path.port | default (.backendApiPort | default 3001) -}}
 {{- $pathType = $path.pathType | default "Exact" -}}
 {{- end -}}
@@ -87,7 +71,7 @@ target: backendApi routes to the main backend API listener instead.
 
 {{/*
 Render an MCP-related HTTPRoute rule. By default rules route to the MCP service;
-target: backendApi routes to the main backend API listener instead.
+target: backendApi routes to the main backend API Service instead.
 */}}
 {{- define "retool.httpRoute.mcpRule" -}}
 {{- $root := .root -}}
@@ -101,7 +85,7 @@ target: backendApi routes to the main backend API listener instead.
 {{- $servicePort := $rule.port | default ($mcpService.externalPort | default 4010) -}}
 {{- $pathType := $rule.pathType | default "PathPrefix" -}}
 {{- if eq $target "backendApi" -}}
-{{- $serviceName = include "retool.fullname" $root -}}
+{{- $serviceName = include "retool.backendApi.name" $root -}}
 {{- $servicePort = $rule.port | default (.backendApiPort | default 3001) -}}
 {{- $pathType = $rule.pathType | default "Exact" -}}
 {{- end -}}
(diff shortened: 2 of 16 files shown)


arnold-retool:
* **#337** <a href="https://app.graphite.com/github/pr/tryretool/retool-helm/337?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/tryretool/retool-helm/337?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

arnold-retool:
@greptile rereview

arnold-retool:
@greptile re-review

charts/retool/templates/deployment_dbconnector.yaml
```diff
@@ -102,7 +102,7 @@ spec:
           - name: DISABLE_DATABASE_MIGRATIONS
             value: "true"
           - name: WORKFLOW_BACKEND_HOST
-            value: http://{{ template "retool.fullname" . }}-workflow-backend
+            value: http://{{ include "retool.workflowBackend.name" . }}
```
arnold-retool, line 105: New helper changes formatting of service specific backend names

arnold-retool: Non-MCP files are touched for shared service-name plumbing, not because their runtime behavior changed.

The chart introduced a common retool.fullnameWithSuffix helper so suffixed service names fit Kubernetes’ 63-character DNS label limit. Once that helper exists, the chart needs internal callers to use the same
helper instead of hand-building names like:

{{ template "retool.fullname" . }}-dbconnector

So these files changed only to keep env vars aligned with the actual generated Service names:

- _workers.tpl: DB_CONNECTOR_HOST / JAVA_DB_CONNECTOR_HOST
- deployment_backend.yaml: db connector and workflow backend host env vars
- deployment_dbconnector.yaml: WORKFLOW_BACKEND_HOST
- deployment_workflows.yaml: db connector host env vars
- _helpers.tpl: adds the shared suffix helper and the new backend-api name helper

For normal short release names, this should render the same hostnames as before. The reason it matters is long release names: without this, a Service name may be truncated one way while env vars still point at the
untruncated hardcoded name. The MCP backend-api Service made that naming consistency more important, so the branch cleaned up the shared helper usage around it.

charts/retool/templates/service.yaml
```diff
@@ -1,9 +1,8 @@
-{{- $mcpNeedsBackendApi := eq (include "retool.mcp.needsBackendApi" .) "true" -}}
+{{- $mcpEnabled := ((.Values.mcp).enabled) | default false -}}
 {{- $backendApiService := (((.Values.mcp).backendMetadata).service) | default dict -}}
-{{- $servicePortName := .Values.service.portName | default "http" -}}
-{{- $backendApiPortName := $backendApiService.portName | default "http-api" -}}
-{{- if and $mcpNeedsBackendApi (eq $servicePortName $backendApiPortName) -}}
-{{- fail "When MCP backend API routing is enabled, .Values.service.portName and .Values.mcp.backendMetadata.service.portName must be different" -}}
+{{- $backendApiServiceEnabled := $mcpEnabled -}}
+{{- if hasKey $backendApiService "enabled" -}}
+{{- $backendApiServiceEnabled = and $mcpEnabled $backendApiService.enabled -}}
```
arnold-retool, line 5: add additional service if both backend and MCP are enabled

charts/retool/templates/_helpers.tpl
```diff
@@ -24,34 +24,18 @@ If release name contains chart name it will be used as a full name.
 {{- end }}
 
 {{/*
-Whether MCP routing needs the main Retool Service to expose the backend API
-listener in addition to the primary frontend-facing port.
+Append a suffix to retool.fullname while preserving the 63-character DNS label
```
arnold-retool, line 27: adds service naming helpers

charts/retool/templates/deployment_mcp.yaml
```diff
@@ -1,11 +1,44 @@
 {{- if .Values.mcp.enabled }}
```
arnold-retool, line 1: spaghetti go templating that can go in _helpers eventually

arnold-retool: cleans domain env var inputs and provides default values

ryanartecona: uhhh since you're adding a bunch of stuff to the top of this file, I think indenting the conditional blocks would be better than leaving them all uniformly left justified. elsewhere in our template files we try to keep conditionals flat when they're intermixed with static yaml since it can get hard to track conditional indentation vs yaml indentation, but a big block of just template stuff like this doesn't have that complication.

charts/retool/templates/service.yaml
ryanartecona: why the new service in this situation? why doesn't just an extra port on the existing backend service work?

ryanartecona (commented):
a few other high level thoughts
* the PR description mentions adding Ingress/HTTPRoute routes that I don't see in the code changes at all. mistake or outdated?
* if I'm reading correctly, the `mcp.oauthMainDomain` value in most cases will just need to be the same as `env.BASE_DOMAIN`
  * if `mcp.oauthMainDomain` is null (default, untouched) can we add a helper that tries to use the value of `env.BASE_DOMAIN` in the spirit of smart defaults? technically the main backend BASE_DOMAIN can come from several different places like `environmentVariables`, `environmentSecrets`, so that won't be 100% coverage, but a valid BASE_DOMAIN setting is already a hard requirement for backend startup these days, and I think putting it in `env` is probably what most people do, so this would ease a config burden.
* I see that `mcp.oauthIntrospectionAuthToken` is still a required option. have you tried to add to the autogenerate-on-first-deploy logic in secret.yaml yet to make that also not required?

charts/retool/templates/service.yaml
arnold-retool: Codex says this ClusterIP service is more secure by default

arnold-retool:
re 1) you're right, outdated

arnold-retool:
re 3), updated secret.yaml

arnold-retool:
re 2), made this defaulting change

charts/retool/templates/service.yaml
ryanartecona: that might be the case, but I don't want to blindly trust robots here. by what logic is it more secure?

arnold-retool: :rolling_on_the_floor_laughing: sorry   
  
The reasoning it gave (that I also believe) is that exposing the 3001 port by default would add a public-facing port, while having this ClusterIP service configuration makes it only available to internal MCP service nodes

charts/retool/templates/_helpers.tpl
```diff
@@ -874,14 +858,21 @@ Usage: {{- include "retool.agentSandbox.backendEnvVars" . | nindent 10 }}
...
 {{- define "retool.mcp.name" -}}
-{{ template "retool.fullname" . }}-mcp
+{{ include "retool.fullnameWithSuffix" (list . "mcp") }}
+{{- end -}}
+
+{{/*
+Set backend API service name
+*/}}
+{{- define "retool.backendApi.name" -}}
+{{ include "retool.fullnameWithSuffix" (list . "backend-api") }}
```
ryanartecona, line 868: now that I understand this is the name for the internal-facing server for mcp to talk to non-public backend endpoints, I think `internal` would be a better name than `backend-api`, as often we colloquially talk about the _existing, public_ server as the "backend api".

charts/retool/templates/secret.yaml
```diff
@@ -50,4 +57,14 @@ data:
...
+
+  {{- if and .Values.mcp.enabled (not $hasOAuthIntrospectionAuthTokenEnv) (not $mcpConfig.oauthIntrospectionAuthTokenSecretName) }}
+  {{- if $mcpConfig.oauthIntrospectionAuthToken }}
+  oauthIntrospectionAuthToken: {{ $mcpConfig.oauthIntrospectionAuthToken | b64enc | quote }}
+  {{- else if (get $secretData "oauthIntrospectionAuthToken") }}
+  oauthIntrospectionAuthToken: {{ get $secretData "oauthIntrospectionAuthToken" }}
+  {{- else }}
+  oauthIntrospectionAuthToken: {{ randAlphaNum 64 | b64enc | quote }}
+  {{- end }}
+  {{- end }}
```
ryanartecona, line 69: nice!

charts/retool/ci/test-mcp-agent-sandbox-secret-option.yaml
```diff
@@ -1,6 +1,11 @@
...
+  agentSandbox:
+    enabled: true
+    externalSecret:
+      name: agent-sandbox-jwt
+
 mcp:
   enabled: true
   config:
+    oauthMainDomain: example.retool.com
     oauthIntrospectionAuthToken: test-oauth-introspection-token
```
ryanartecona, line 11: probably want to remove this one now too, or possibly both of these, for the CI values if we want it to be autogenerated/derived in most cases?

you might have already seen that these `ci/test-...-option.yaml` values get matrixed and layered on top of the other `ci/...-values.yaml` file when validating the whole helm template render in CI. I think the most important point is that there's 1 option file that has the simplest/common case, i.e. just a `mcp: enabled: true` and no other config, and then 1 option file that actually sets values for most/all of the other configs available (or at least the ones you care to test). the disabled case will already be covered by the other unrelated option files which don't enable mcp.

ryanartecona (approved):
still a couple suggestions but we're getting close enough here that I'll preemptively approve

arnold-retool:
### Merge activity

* **Jun 24, 9:46 PM UTC**: @arnold-retool merged this pull request with [Graphite](https://app.graphite.com/github/pr/tryretool/retool-helm/337).

PR #338, opened 2026-06-19
rr: git server inherits rr.enabled; fail on bad blob storage (warn if skip set)

JatinNanda:
## what

Two changes to `rr.gitServer` in `charts/retool`:

1. **gitServer inherits the `rr.enabled` master switch** — same tri-state as `jsExecutor`/`agent`/`agentSandbox`. New `retool.gitServer.enabled` helper backed by `retool.rr.componentEnabled`; every gitServer gate routes through it; `rr.gitServer.enabled` defaults to `null` (inherit). Explicit `true|false` still overrides.

2. **Hybrid blob-storage validation.** When the git server is enabled and blob storage is misconfigured (not exactly one of `rr.blobStorage.s3/gcs/azure`, and no `RR_BLOB_STORAGE_PROVIDER`/`RR_DEFAULT_*` supplied directly):
   - **default → hard fail** (render aborts; blocks the deploy)
   - **`rr.gitServer.skipBlobStorageValidation: true` → non-fatal NOTES.txt warning** — the escape hatch for providers injected via `envFrom` that can't be inspected at template time.

## why

`rr.gitServer.enabled` defaulted to `false` and did not follow the master switch, so enabling rr without explicitly enabling gitServer left React Retool git silently broken. This makes it inherit. Validation stays fail-fast by default but can be downgraded to a warning when the operator knows the provider is supplied another way.

## behavior change

Consumers with `rr.enabled: true` who did not explicitly set `gitServer` now get the git server, and the render fails if blob storage isn't configured (unless they set `skipBlobStorageValidation`). Consumers that set `rr.gitServer.enabled` explicitly are unaffected.

## caveat

The skip-path warning shows in `helm install`/`upgrade`/`--dry-run` output, not in `helm template`/CI renders.

## verification

- `helm lint` passes; existing `git-server-separate` CI fixture still renders.
- gitServer inherits `rr.enabled`; explicit `gitServer.enabled: false` overrides.
- misconfigured (zero or 2+ providers), skip=false → render FAILS; skip=true → renders + warning; configured provider → renders clean.

### DIFF
--- a/charts/retool/templates/_helpers.tpl
+++ b/charts/retool/templates/_helpers.tpl
@@ -453,6 +453,15 @@ Usage: (include "retool.agent.enabled" .)
 {{- include "retool.rr.componentEnabled" (dict "root" . "component" "agent") -}}
 {{- end -}}
 
+{{/*
+Set RR git server enabled. Honors the shared RR master switch (rr.enabled)
+unless rr.gitServer.enabled is set explicitly true|false.
+Usage: (include "retool.gitServer.enabled" .)
+*/}}
+{{- define "retool.gitServer.enabled" -}}
+{{- include "retool.rr.componentEnabled" (dict "root" . "component" "gitServer") -}}
+{{- end -}}
+
 {{/* Global Temporal configuration */}}
 {{- define "retool.temporalConfig" -}}
 {{- .Values.workflows.temporal | default .Values.temporal | toYaml -}}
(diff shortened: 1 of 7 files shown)


charts/retool/templates/_helpers.tpl
```diff
@@ -975,17 +984,17 @@ directly via environmentVariables / environmentSecrets).
...
-is enabled. Skipped when the user has plumbed the RR_BLOB_STORAGE_PROVIDER /
-RR_DEFAULT_*_* env vars in directly via env/environmentVariables/environmentSecrets,
-which is treated as an opt-out from the first-class blobStorage config.
-Also skipped entirely when rr.gitServer.skipBlobStorageValidation is true, which
-is the escape hatch for sources we cannot inspect at template time (e.g. env
-vars injected via envFrom from a Secret/ConfigMap).
-No-op when rr.gitServer is disabled.
+Returns "1" when rr.gitServer is enabled but no blob-storage provider is
+configured -- i.e. a misconfiguration. Surfaced as a NON-FATAL warning in
+NOTES.txt (git_server still renders, but git ops fail at runtime without it).
```
emas-rt: why non-fatal?

JatinNanda: made fatal, but non-fatal if the skip flag is on

ryanartecona (approved):
LGTM

PR #344, opened 2026-06-29
[fix][MCP] Wire OAUTH_INTROSPECTION_AUTH_TOKEN into the backend

justinpchen94:
## Summary

When the MCP server is enabled on a self-hosted Helm deployment, it can get stuck registering only the 4 read-only tools because OAuth token introspection fails. The MCP server calls `/api/oauth2/introspect` on the main Retool backend, and the backend rejects that call unless `OAUTH_INTROSPECTION_AUTH_TOKEN` is set to the same value MCP sends.

The chart only injected `OAUTH_INTROSPECTION_AUTH_TOKEN` into the **MCP deployment** (`deployment_mcp.yaml`), never the **backend**. So the backend logged `OAuth token introspection is not configured: OAUTH_INTROSPECTION_AUTH_TOKEN is missing`, the MCP server logged `Falling back to legacy MCP auth behavior after token introspection failure`, and only the read-only toolset registered. This was undiscoverable: the `mcp.config.oauthIntrospectionAuthToken` docs gave no hint the backend needed the same value, and the chart-owned auto-generate path made it worse by putting a token on MCP that the backend could never match.

Reported by a customer on a v4.0.3-stable / chart v6.11.6 deployment built from the `values.yaml` template.

## What changed

- `deployment_backend.yaml`: when `mcp.enabled`, inject `OAUTH_INTROSPECTION_AUTH_TOKEN` into the backend from the **same source** the MCP deployment uses, in the same precedence order:
  1. `mcp.config.oauthIntrospectionAuthTokenSecretName` → `secretKeyRef`
  2. `mcp.config.oauthIntrospectionAuthToken` → literal `value`
  3. chart-owned config secret → `secretKeyRef` into the release secret's `oauthIntrospectionAuthToken` key
  
  This guarantees backend and MCP always agree. It is skipped when the operator already wired the token onto the backend (`.Values.env`, `.Values.environmentVariables`, or `.Values.environmentSecrets`), so the existing global-`env:` workaround keeps working with no duplicate env entry. All three branches are also skipped when the token is supplied via `mcp.environmentVariables`: the MCP deployment uses that value over its config-based branches, so auto-wiring the backend from a config source could inject a differing value and silently break introspection. In that case the backend is not auto-wired; set `OAUTH_INTROSPECTION_AUTH_TOKEN` in the global `env:` block.
- `values.yaml` / `charts/retool/values.yaml`: document that the chart now wires the token into the backend automatically, and note the case where the token is set via `mcp.environmentVariables` and you must also set it in the global `env:` block.
- `Chart.yaml`: bump `6.11.6` → `6.11.7`.

## Test Plan

Rendered `helm template` across the token-source scenarios and confirmed the backend env matches what the MCP deployment uses:

| Scenario | Backend result |
| --- | --- |
| Chart-owned auto-generated secret | `secretKeyRef` → release secret / `oauthIntrospectionAuthToken` (identical to MCP) |
| `mcp.config.oauthIntrospectionAuthToken` literal | `value:` matches MCP literal |
| `mcp.config.oauthIntrospectionAuthTokenSecretName` (+ externalSecrets) | `secretKeyRef` → named secret / key |
| Token only via `mcp.environmentVariables` | no backend entry, no dangling secret ref |
| `mcp.environmentVariables` + `mcp.config.oauthIntrospectionAuthTokenSecretName` | no backend entry (MCP uses the env value; backend not auto-wired from config, avoiding a mismatch) |
| Operator set `env.OAUTH_INTROSPECTION_AUTH_TOKEN` + `mcp.config` token | exactly one backend entry, from `env` (no double-inject) |

`helm lint` passes; both `values.yaml` files remain byte-identical (sync check).

### DIFF
--- a/charts/retool/Chart.yaml
+++ b/charts/retool/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 6.11.6
+version: 6.11.7
 maintainers:
   - name: Retool Engineering
     email: engineering+helm@retool.com
--- a/charts/retool/templates/deployment_backend.yaml
+++ b/charts/retool/templates/deployment_backend.yaml
@@ -312,13 +312,13 @@ spec:
           {{- end }}
           {{- end }}
           {{- if not $backendHasIntrospectionToken }}
-          {{- if $mcpConfig.oauthIntrospectionAuthTokenSecretName }}
+          {{- if and $mcpConfig.oauthIntrospectionAuthTokenSecretName (not $mcpEnvHasIntrospectionToken) }}
           - name: OAUTH_INTROSPECTION_AUTH_TOKEN
             valueFrom:
               secretKeyRef:
                 name: {{ $mcpConfig.oauthIntrospectionAuthTokenSecretName }}
                 key: {{ $mcpConfig.oauthIntrospectionAuthTokenSecretKey | default "oauthIntrospectionAuthToken" }}
-          {{- else if $mcpConfig.oauthIntrospectionAuthToken }}
+          {{- else if and $mcpConfig.oauthIntrospectionAuthToken (not $mcpEnvHasIntrospectionToken) }}
           - name: OAUTH_INTROSPECTION_AUTH_TOKEN
             value: {{ $mcpConfig.oauthIntrospectionAuthToken | quote }}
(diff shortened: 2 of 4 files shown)


justinpchen94:
@greptile

rsuresh-retool (approved):

slaminad (approved):

PR #351, opened 2026-07-23
[docs][MCP] Document simplified ingress for Retool 4.0.7+

arnold-retool:
## Summary

- document that Retool 4.0.7+ can route `/mcp`, OAuth discovery, and normal application traffic through the main Retool Service on port 3000
- document the required backend setting `MCP_SERVICE_INGRESS_DOMAIN=http://<fullname>-mcp:4010`
- retain the explicit 3001/4010 path mapping for Retool versions before 4.0.7

## Testing

- `helm lint charts/retool`
- `helm lint charts/retool -f charts/retool/ci/test-mcp-enabled-option.yaml`
- rendered the simplified setup with the MCP-specific Ingress and HTTPRoute rules disabled
- verified the backend receives `MCP_SERVICE_INGRESS_DOMAIN` and no public `/mcp` path is rendered
- verified `values.yaml` and `charts/retool/values.yaml` remain identical
- `git diff --check`

### DIFF
--- a/charts/retool/Chart.yaml
+++ b/charts/retool/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 6.11.12
+version: 6.11.13
 maintainers:
   - name: Retool Engineering
     email: engineering+helm@retool.com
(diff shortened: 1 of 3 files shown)


charts/retool/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 6.11.12
+version: 6.11.13
```
ryanartecona: I know CI will complain at you but you don't need to bump the patch version for this PR if it will only be a purely comment change

ryanartecona (approved):

PR #352, opened 2026-07-24
[chore][MCP] Configure backend MCP service ingress domain

arnold-retool:
## Summary

- Inject MCP_SERVICE_INGRESS_DOMAIN into the main backend when MCP is enabled.
- Derive the internal URL from the release-specific MCP Service name and configured external port.
- Preserve explicit overrides from env, environmentSecrets, or environmentVariables.
- Update both synchronized values.yaml files to document the automatic configuration.

## Testing

- helm lint charts/retool
- helm template with test-install-values.yaml and test-mcp-enabled-option.yaml
- Verified MCP-disabled, default, custom Service name/port, and explicit override renders
- Confirmed values.yaml and charts/retool/values.yaml remain synchronized
- git diff --check

### DIFF
--- a/charts/retool/templates/_helpers.tpl
+++ b/charts/retool/templates/_helpers.tpl
@@ -52,6 +52,26 @@ env.BASE_DOMAIN. Secret-backed values cannot be resolved at template time.
 {{- end }}
 
 {{/*
+MCP Service env var for the main Retool backend. Explicit backend env settings
+take precedence over the chart-generated in-cluster Service URL.
+*/}}
+{{- define "retool.mcp.backendEnvVars" -}}
+{{- if .Values.mcp.enabled }}
+{{- $backendHasMcpServiceIngressDomain := hasKey (.Values.env | default dict) "MCP_SERVICE_INGRESS_DOMAIN" -}}
+{{- range .Values.environmentVariables }}
+{{- if eq .name "MCP_SERVICE_INGRESS_DOMAIN" }}
+{{- $backendHasMcpServiceIngressDomain = true }}
+{{- end }}
+{{- end }}
+{{- if not $backendHasMcpServiceIngressDomain }}
+{{- $mcpService := .Values.mcp.service | default dict }}
+- name: MCP_SERVICE_INGRESS_DOMAIN
+  value: {{ printf "http://%s:%v" (include "retool.mcp.name" .) ($mcpService.externalPort | default 4010) | quote }}
+{{- end }}
+{{- end }}
+{{- end -}}
+
+{{/*
 Render an MCP-related Ingress path. By default paths route to the MCP service.
 target: backendInternal routes to the backend API Service.
 */}}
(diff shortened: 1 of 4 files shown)


charts/retool/templates/_helpers.tpl
```diff
@@ -51,6 +51,31 @@ env.BASE_DOMAIN. Secret-backed values cannot be resolved at template time.
...
 {{- end }}
 
+{{/*
+MCP Service env var for the main Retool backend. Explicit backend env settings
+take precedence over the chart-generated in-cluster Service URL.
+*/}}
+{{- define "retool.mcp.backendEnvVars" -}}
+{{- if .Values.mcp.enabled }}
+{{- $backendHasMcpServiceIngressDomain := hasKey (.Values.env | default dict) "MCP_SERVICE_INGRESS_DOMAIN" -}}
+{{- range .Values.environmentSecrets }}
```
lukefoster11: Not sure if we need to check all these different places for MCP_SERVICE_INGRESS_DOMAIN being defined. I think we can narrow these checks down to just .Values.env.
Also wondering if there's ever even a case where the MCP_SERVICE_INGRESS_DOMAIN shouldn't just be set to `value: {{ printf "http://%s:%v" (include "retool.mcp.name" .) ($mcpService.externalPort | default 4010) | quote }}`? Should we even allow overriding this default?

arnold-retool: - Just `.Values.env` could make setup less straightforward for users with `.environmentSecrets` and `.environmentVariables` in their charts.
- An override is generally useful in cases with custom networking. I can't imagine users taking advantage of this override to break things for themselves.

Thanks for the comment! Let me know if you feel strongly about either of these.

lukefoster11 (approved):
approving but with comment that I think we can still get rid of the check within environmentSecrets. I don't feel super strongly about it, but I think we should reserve that section for actual secrets

arnold-retool:
Oh, put that way, I think that makes sense. I'll make that revision

arnold-retool:
* **#352** <a href="https://app.graphite.com/github/pr/tryretool/retool-helm/352?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/tryretool/retool-helm/352?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

arnold-retool:
@greptile re review with revision removing secret check

arnold-retool:
### Merge activity

* **Aug 4, 6:18 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/tryretool/retool-helm/352).
* **Aug 4, 6:18 PM UTC**: @arnold-retool merged this pull request with [Graphite](https://app.graphite.com/github/pr/tryretool/retool-helm/352).

PR #359, opened 2026-08-07
[fix][r2][SHI-458] Fix rr.gitServer.separate.enabled=true: blob storage env vars

emas-rt:
## Summary

Fixes a bug affecting self-hosted deployments using `rr.gitServer.separate.enabled=true`.

### SHI-458 — Backend missing blob storage env vars ([Linear](https://linear.app/retool/issue/SHI-458))

`retool.gitServer.commonEnv` (which emits `RR_BLOB_STORAGE_PROVIDER` / `RR_DEFAULT_S3_*` / `RR_DEFAULT_GCS_*` / `RR_DEFAULT_AZURE_*`) was gated inside the `else` branch of the `separateEnabled` check in `deployment_backend.yaml`. When the git server was separate, only `RR_GIT_SERVER_HOST` and `RR_GIT_SERVER_PORT` were set on the backend — causing `/api/ai/rr/snapshots` to fail with `Required environment variable RR_DEFAULT_S3_BUCKET is not set`.

**Fix:** Move `retool.gitServer.commonEnv` outside the if/else so it renders unconditionally whenever `gitServer.enabled=true`, in both embedded and separate modes.

## Test plan

- [x] `helm template` with `rr.gitServer.separate.enabled=true` + S3 blob storage: verify `RR_GIT_SERVER_HOST`, `RR_GIT_SERVER_PORT`, and `RR_BLOB_STORAGE_PROVIDER`/`RR_DEFAULT_S3_*` all appear on the backend deployment
- [x] `helm template` with `rr.gitServer.separate.enabled=false` (embedded): verify no `RR_GIT_SERVER_HOST` on backend, blob storage vars still render
- [x] End-to-end: deployed with `rr.gitServer.separate.enabled=true` and `rr.blobStorage.s3` configured — `/api/ai/rr/snapshots` returns 201

### DIFF
--- a/charts/retool/Chart.yaml
+++ b/charts/retool/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 6.11.15
+version: 6.11.16
 maintainers:
   - name: Retool Engineering
     email: engineering+helm@retool.com
--- a/charts/retool/templates/deployment_backend.yaml
+++ b/charts/retool/templates/deployment_backend.yaml
@@ -273,9 +273,8 @@ spec:
             value: {{ template "retool.gitServer.name" . }}
           - name: RR_GIT_SERVER_PORT
             value: {{ include "retool.gitServer.port" . | quote }}
-          {{- else }}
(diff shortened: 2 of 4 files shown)


lukefoster11 (approved):
👍

PR #360, opened 2026-08-11
[SHI-448] Option for code executor image tags that don't match backend

justinyang-retool:
Customer making a Helm Chart update had with the one-off image they were using `tryretool/code-executor-one-offs:pr-3.334.6-fcf7211`.  The image name does not match the backend image which did not include the pr-.  By design in the helm chart, we don't allow for differing image names between backend and code executor.  There should be an option for the customer here to be able to override the image for codeExecutor in this case, or we need to make sure the names match.

Note I've decided for now that pullpolicy will not be overrideable

Testing:
In-code testing:
- Put through code a template file:
Testing with no override:
Screenshot 2026-08-12 at 8 39 21 PM
Testing with override:
Screenshot 2026-08-12 at 8 40 43 PM

Deploy testing:
Before case:
Screenshot 2026-08-13 at 6 46 13 PM
has the default image and the image pulled successfully
Also note the code executor has no image tag while the backend image tag is the one pulled:
Screenshot 2026-08-13 at 6 48 42 PM
Screenshot 2026-08-13 at 6 49 42 PM

After case:
Screenshot 2026-08-13 at 7 31 28 PM
Has the new image for code executor
Note that the code executor pulled successfully
THe pod `justinyang-test-547579fb56-nws4p` is erroring out but this actually precedes the image override update, and it uses the default image so it's not affected
Also note the helm chart:
Screenshot 2026-08-13 at 7 31 50 PM
Screenshot 2026-08-13 at 7 31 56 PM


charts/retool/ci/test-code-executor-image-tag-option.yaml
```diff
@@ -0,0 +1,20 @@
...
+# the chart default and must behave the same as absent, so an operator who
+# leaves the key in place (or copies it from values.yaml) still inherits instead
+# of rendering an empty tag.
+# The tag must stay above the chart's code-executor and workflows version gates
+# (>= 3.20.15 and >= 3.6.11), or the deployment is not rendered at all and the
+# test silently asserts nothing.
+image:
+  tag: 9.9.9-test
+
+codeExecutor:
```
justinyang-retool: TODO: default should not set image at all here

charts/retool/ci/test-code-executor-image-override-option.yaml
```diff
@@ -0,0 +1,21 @@
...
+# *values.yaml by .github/kubeconform.sh.
+#
+# Pairs with test-code-executor-image-tag-option.yaml, which sets the same
+# top-level tag but leaves the override unset. Rendering both makes the
+# difference explicit.
+#
+# The repository and tag below are deliberately unlike the top-level image.*, so
+# a regression that inherits the backend values, or that falls back to the
+# chart's built-in default tag, changes the rendered image and fails the test.
+# The tag must stay above the chart's code-executor and workflows version gates
```
justinyang-retool, line 8: TODO: current test fixtures test nothing, this is just dummy code
Not sure if its worth adding those, or if i should just do a manual deploy test and skip test code

charts/retool/values.yaml
```diff
@@ -781,11 +781,15 @@ mcp:
...
+  # Code executor normally tracks the top-level `image` parameters. Override the
+  # tag only when you need a one-off or patched build (e.g. a support build whose
+  # tag differs from the backend's); running a code executor version that does not
+  # match the backend is not generally supported.
 
   image:
     repository: tryretool/code-executor-service
+    # defaults to top level image.tag
+    tag: null
     pullPolicy: IfNotPresent
```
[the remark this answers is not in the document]
justinyang-retool: TODO: I didn't actually update any logic to check for pullpolicy override anyways since its barely used
so maybe ill just remove the field entirely (not sure if its allowed, will have to check semantics)

charts/retool/ci/test-code-executor-image-tag-option.yaml
```diff
@@ -0,0 +1,20 @@
...
+  image:
+    tag: null
```
ryanartecona: we don't need this file, since the other tests already have a null codeExecutor image (by default) so this is matrixing a redundant check

charts/retool/ci/test-code-executor-image-override-option.yaml
justinyang-retool: @ryanartecona What is your understanding of the test fixture system here? My testing mostly focused on deploying to a balloon instance, so I didn't actually use these test fixtures at all 😅

charts/retool/values.yaml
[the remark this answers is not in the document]
ryanartecona: reminder to resolve this one. I don't have strong feelings whether we have codeExecutor.image.pullPolicy or not, but we shouldn't leave it in the values.yaml defaults file if it doesn't actually do anything.

charts/retool/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 6.11.16
+version: 6.12.0
```
ryanartecona: not worth a minor version IMO, as it's just a new escape hatch to unblock niche situations or hardened image users

```suggestion
version: 6.11.17
```

charts/retool/values.yaml
ryanartecona: please remove the pullPolicy from the default values.yaml as it doesn't do anything

values.yaml
```diff
@@ -781,11 +781,15 @@ mcp:
...
+  # Code executor normally tracks the top-level `image` parameters. Override the
+  # tag only when you need a one-off or patched build (e.g. a support build whose
+  # tag differs from the backend's); running a code executor version that does not
+  # match the backend is not generally supported.
 
   image:
     repository: tryretool/code-executor-service
+    # defaults to top level image.tag
+    tag: null
     pullPolicy: IfNotPresent
```
ryanartecona: same here, please remove the pullPolicy since it's dead code and misleading

ryanartecona (approved):

charts/retool/ci/test-code-executor-image-tag-option.yaml
```diff
@@ -0,0 +1,15 @@
...
+    # A one-off / support build lives in a different repository than the backend
+    # and carries a tag the backend image does not have (SHI-448).
+    repository: tryretool/code-executor-one-offs
+    # Deliberately differs from the top-level image.tag so the rendered
+    # deployment proves the per-workload tag override wins over the backend tag.
+    tag: pr-3.334.6-fcf7211
+    # Deliberately differs from the global image.pullPolicy (IfNotPresent in the
+    # base values) so the rendered deployment proves the per-workload override is
+    # honored rather than the global value.
+    pullPolicy: Always
```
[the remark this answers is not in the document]
justinyang-retool: The existing pattern does not perform these assertions. The test I added should only need to pass the existing checks as per existing patterns, that's it.

PR #363, opened 2026-08-13
[INF-7324] (opt-in) hpa for jse deployment

lukefoster11:
Easy to enable hpa for the js-executor deployment. Scale on 70% mem and 70% cpu usage.
Ask from Colgate

### DIFF
--- a/charts/retool/Chart.yaml
+++ b/charts/retool/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 6.11.18
+version: 6.11.19
 maintainers:
   - name: Retool Engineering
     email: engineering+helm@retool.com
--- a/charts/retool/ci/test-js-executor-hpa-option.yaml
+++ b/charts/retool/ci/test-js-executor-hpa-option.yaml
@@ -1,8 +1,8 @@
 rr:
 
   # Exercises the JS executor HPA (opt-in via rr.jsExecutor.autoscaling).
-  # Overlaid on top of test-install-values.yaml. Covers CPU, memory, extra
-  # metrics, and scale behavior branches; replicaCount must be omitted from
+  # Overlaid on top of test-install-values.yaml. Covers CPU and memory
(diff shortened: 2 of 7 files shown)


lukefoster11:
@greptile review me again

lukefoster11:
@greptile review me again pls

lukefoster11:
@greptile can you review again

lukefoster11:
@greptile review me again pls

init-b (approved):
LGTM! just wondering what happens to JS executions when it scales down?

lukefoster11:
> LGTM! just wondering what happens to JS executions when it scales down?

The replicacount will just go down to the min replica count (1 by default but configurable). JSE shouldn't hit performance issues while cpu and mem utilization are below 70% so JS execution speed shouldn't be affected. One thing I just realized is that this HPA depends on having metrics-server enabled, so I will add a comment that if that isn't enabled then the HPA won't work properly and will stay at the min replica count.

init-b:
> > LGTM! just wondering what happens to JS executions when it scales down?
> 
> The replicacount will just go down to the min replica count (1 by default but configurable). 

@lukefoster11 any implications about the in-flight process in the container we kill? I really don't know how that currently works in our Cloud environment.

lukefoster11:
> > > LGTM! just wondering what happens to JS executions when it scales down?
> > 
> > 
> > The replicacount will just go down to the min replica count (1 by default but configurable).
> 
> @lukefoster11 any implications about the in-flight process in the container we kill? I really don't know how that currently works in our Cloud environment.

It's a graceful termination, same as when we terminate all jse pods when we restart a deployment. In-flight requests complete before the pod spins down

init-b:
> > > > LGTM! just wondering what happens to JS executions when it scales down?
> > > 
> > > 
> > > The replicacount will just go down to the min replica count (1 by default but configurable).
> > 
> > 
> > @lukefoster11 any implications about the in-flight process in the container we kill? I really don't know how that currently works in our Cloud environment.
> 
> It's a graceful termination, same as when we terminate all jse pods when we restart a deployment. In-flight requests complete before the pod spins down

Thanks for the context Luke!

PR #364, opened 2026-08-16
[fix][plat] Allow for npm registry env vars to be set

jamie-retool:
Adds easy setting of npm registry settings for the agent executor proxy.

### DIFF
--- a/charts/retool/templates/_helpers.tpl
+++ b/charts/retool/templates/_helpers.tpl
@@ -783,6 +783,9 @@ or the catch-all externalSecret.name. No-op when agentSandbox is disabled.
 {{- if and $npm.authTokenSecret.name (not $npm.authTokenSecret.key) -}}
 {{- fail "agentSandbox.proxy.npmRegistry.authTokenSecret.name is set without .key, which would render an empty secretKeyRef and leave the proxy unable to start. Set .key to the entry holding the registry token." -}}
 {{- end -}}
+{{- if and $npm.authToken $npm.authTokenSecret.name -}}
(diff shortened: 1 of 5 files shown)


charts/retool/templates/_helpers.tpl
```diff
@@ -775,6 +775,18 @@ or the catch-all externalSecret.name. No-op when agentSandbox is disabled.
...
 {{- fail "agentSandbox.enabled requires an encryption key: the proxy derives the sandbox-iframe asset-token HMAC key from it and throws when serving a sandbox without it, and the backend must use the same value. Set agentSandbox.encryptionKey (64 hex chars, openssl rand -hex 32) or agentSandbox.externalSecret.name (with an encryption-key entry)." -}}
 {{- end -}}
+{{- $npm := $as.proxy.npmRegistry -}}
+{{- if not (has $npm.authMode (list "bearer" "none")) -}}
+{{- fail (printf "agentSandbox.proxy.npmRegistry.authMode must be \"bearer\" or \"none\" (got %q). The proxy rejects any other value at startup." $npm.authMode) -}}
+{{- end -}}
+{{- if $npm.url -}}
+{{- if and $npm.authTokenSecret.name (not $npm.authTokenSecret.key) -}}
+{{- fail "agentSandbox.proxy.npmRegistry.authTokenSecret.name is set without .key, which would render an empty secretKeyRef and leave the proxy unable to start. Set .key to the entry holding the registry token." -}}
+{{- end -}}
```
ryanartecona, line 785: maybe worth another check here to fail if `$npm.authToken` and `$npm.authTokenSecret.name` are both set, since the secret won't get used at all in that case?

ryanartecona (approved):
LGTM

PR #365, opened 2026-08-20
[chore][PLAT-1440] Raise termination grace periods so in-flight SLFs can finish

nataliechen-pixel:
Related:
- Cloud / MSH drain: https://github.com/tryretool/retool-k8s/pull/18842
- Helm (self-hosted): https://github.com/tryretool/retool-helm/pull/365
- JSE process shutdown: https://github.com/tryretool/retool_development/pull/83807

Standardize and increase `terminationGracePeriodSeconds` across backend, dbconnector, codeExecutor, and jsExecutor deployments to better accommodate in-flight requests during pod shutdown.

### What changed?

- `backend` `terminationGracePeriodSeconds` increased from 136s (2m 15s + 1s) to 391s (6m 30s + 1s)
- `dbconnector` `terminationGracePeriodSeconds` changed from a commented-out 960s default to an active value of 391s
- `codeExecutor` now has an explicit `terminationGracePeriodSeconds` of 391s (6m in-flight + 30s preStop buffer + 1s), previously unset
- `jsExecutor` now has an explicit `terminationGracePeriodSeconds` of 391s (6m in-flight SLF + 30s drain buffer + 1s), previously unset

### How to test?

Deploy the chart and verify that pods for the backend, dbconnector, codeExecutor, and jsExecutor reflect the updated `terminationGracePeriodSeconds` values in their pod specs. During a rolling update, confirm that in-flight requests are allowed to complete before pods are forcibly terminated.

### Why make this change?

The previous default grace period for the backend (136s) was based on a 2-minute query timeout, which no longer reflects the actual in-flight request duration. Aligning all services to a 6-minute baseline (plus appropriate drain buffers) ensures pods have sufficient time to finish processing requests before Kubernetes forcibly kills them, reducing disruption during updates and rescheduling events.

### DIFF
--- a/charts/retool/templates/deployment_backend.yaml
+++ b/charts/retool/templates/deployment_backend.yaml
@@ -46,7 +46,7 @@ spec:
 {{ toYaml .Values.ui.labels | indent 8 }}
 {{- end }}
     spec:
-      terminationGracePeriodSeconds: {{ .Values.backend.terminationGracePeriodSeconds | default 136 }}
+      terminationGracePeriodSeconds: {{ .Values.backend.terminationGracePeriodSeconds | default 391 }}
       serviceAccountName: {{ template "retool.serviceAccountName" . }}
       {{- if .Values.priorityClassName }}
       priorityClassName: "{{ .Values.priorityClassName }}"
--- a/charts/retool/Chart.yaml
+++ b/charts/retool/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: retool
 description: A Helm chart for Kubernetes
 type: application
-version: 6.11.20
+version: 6.11.21
 maintainers:
   - name: Retool Engineering
     email: engineering+helm@retool.com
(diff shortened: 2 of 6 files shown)


charts/retool/templates/deployment_backend.yaml
```diff
@@ -46,7 +46,7 @@ spec:
 {{ toYaml .Values.ui.labels | indent 8 }}
 {{- end }}
     spec:
-      terminationGracePeriodSeconds: {{ .Values.backend.terminationGracePeriodSeconds | default 136 }}
```
mako-taco, line 49: is it possible to add comments in the YAML to explain how all of these magic numbers were arrived at?

mako-taco: ah I see... they're documented in the default values. Do we need the `|` still, considering we have default values checked in?

mako-taco (approved):

